Issue #357, opened 2014-09-11
gud integration


birkenfeld:
GUD has a nice interface to pdb, but I don't know if it's widely used because it's kind of hard to find. (Also, `pdb` defaults to looking for an executable named "pdb", while nowadays one should use "python -m pdb".)

elpy could make a few entry functions available that make starting gud-pdb easier, e.g.
- run this file with pdb
- run "something" (to be queried in minibuffer) with pdb and set a breakpoint at the current line

azzamsa:
any update on this ?

galaunay:
That's a pretty old one, but still a good idea I think.

We could easily provide the following functions relying on GUD:
- `elpy-gud-debug-buffer`
- `elpy-gud-debug-buffer-and-break-at-point`

But IMHO, because `python.el` support quite well pdb, it may be better to provide a bunch of functions to allow debugging directly in the interpreter, like:
- `elpy-pbd-debug-buffer`
- `elpy-pbd-debug-buffer-and-break-at-point`
- `elpy-pdb-debug-last-exception`

azzamsa:
yes, that wiil be great.

with `elpy-pbd-debug-buffer` and `elpy-pbd-debug-buffer-and-break-at-point` we don't have to have more step to do debugging.

PR #370, opened 2014-10-07
Add debugging functions.

ionrock:
This will run pdb, set a break point and clear all the breakpoints.

I haven't set any keybindings or implemented a history look up as it seems worthwhile to see if this is something others use. 

Let me know what should be changed / updated. Also, if this is OK to merge but you'd like to see some feature added, please feel free to give me a ticket. 

Thanks for taking a look!

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -2216,6 +2216,38 @@ See http://debbugs.gnu.org/cgi/bugreport.cgi?bug=17647"
                                  end-time)))
       (accept-process-output process timeout))))
 
+;;;;;;;;;;;;;;;;
+;;; Debugging
+
+
+(defun elpy-debug (arg)
+  "Run the current file using pdb."
+  (interactive "P")
+  (let* ((cmd (if (not arg)
+		  (concat "python -m pdb " (buffer-file-name))
+		(read-string "Script/Command: "))))
+    (pdb cmd)))
+
+(defun elpy-debug-insert-break-point ()
+  "Insert a pdb set_trace breakpoint."
+  (interactive)
+  (save-excursion
+    (move-beginning-of-line 1)
+    (python-indent-line)
+    (insert "import pdb; pdb.set_trace();\n")
+    (python-indent-line)))
+
+(defun elpy-debug-clear-all-break-points ()
+  "Remove all pdb set_trace breakpoints.
+
+This specifically looks for breakpoints in the form of
+'import pdb; pdb.set_trace();\n'."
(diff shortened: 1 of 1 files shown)


jorgenschaefer:
I am sorry for having left this lying around for so long without any further response.

First of all, again, thank you for the contribution. Sadly, I am unsure what to do with it, exactly. And I do feel bad for taking so long only to come up with a mostly-negative response.

There are three commands in this patch.

The first one is a command to run the current file under a debugger. This is useful, but I am not sure if this is a good implementation. Quite regularly, I am working in a module somewhere. To run this under a debugger, I need to run the _main program_ under pdb, not the current module. Elpy currently lacks a way of specifying this main program. So this command would, for the most part, not add anything over `pdb` for me. So I am unsure if I want to include it as is.

In my little imaginary world, `elpy-debug` would run "the current program" under pdb. This would require a way of specifying "the current program". Then we could also have a command that simply _executes_ the "current program", instead of running it under a debugger. I am unsure how to do this well, or how the interface for specifying "the current program" would look like. Many projects have more than one, so it would require some sort of election. Tricky!

The second command adds a pdb breakpoint at point. This is currently handled by yasnippet simply by typing `pdb<TAB>` anywhere in the buffer. I do not think a command would simplify this much.

Leaves the command to remove pdb lines. I like that, actually. Can you think of a good key binding for this? `C-c C-b r` might work, with `C-c C-b` a general debugging prefix?

ionrock:
No worries! I forgot I submitted this :)

I sat with this for a little while and came to a similar conclusion. It was difficult to use it consistently.

The `C-c C-b` seems fine to me. That said, if you choose to skip it completely, that is OK too. Thanks for taking a look!

whirm:
What about integrating something like realgud? https://github.com/rocky/emacs-dbgr (which supports setting breakpoints without having to modify the source code and other cool stuff) and using a dir-local to store the different execution profiles?

birkenfeld:
As I said in https://github.com/jorgenschaefer/elpy/issues/357 I tested it in September and thought emacs-dbgr is not ready yet.  But it seems to be actively developed, so it'll have to be checked again.

whirm:
Oh, I had completely forgot about that, sorry :D

jorgenschaefer:
Hm. I keep pushing this forward. I don't think this is useful enough to be included in Elpy, so I will simply close this. Despite of this, I do greatly appreciate the suggestion!

PR #416, opened 2014-11-21
Test commands do not need quote escapes on Windows

fleimgruber:
Another one for Windows compatibility. In essence, on Windows we do not need to escape quotes in the test run command. Test commands started from elpy will fail.

My approach was to `let` bind the correct function depending on OS and leave the mapping in. For Windows the `identity` function is mapped.

The Travis CI results show that this does not work on Emacs version 24.3. Any ideas why it is just failing for that version?

Also, I was trying to add a test for this, but found no `test-quote-escape-in-test-run` or similar in the test suite. Also, I do not quite know how I would go about to test this function. Should we refactor the quote escape mapping to a separate function that just returns the test command string (so we can test it separately)?

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -1827,8 +1827,11 @@ This uses the `elpy-test-runner-p' symbol property."
 
 (defun elpy-test-run (working-directory command &rest args)
   "Run COMMAND with ARGS in WORKING-DIRECTORY as a test command."
-  (let ((default-directory working-directory))
-    (compile (mapconcat #'shell-quote-argument
+  (let ((default-directory working-directory)
+        (quote-fun (if (eq system-type 'windows-nt)
+                       #'identity
+                     #'shell-quote-argument)))
+    (compile (mapconcat quote-fun
                         (cons command args)
                         " "))))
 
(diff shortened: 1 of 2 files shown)


jorgenschaefer:
I just noticed that github does not show PRs in the issue list when I check for the latest release. Great. Sorry for neglecting this for so long.

I suspect this does not require merging, right? It's a bug in Emacs?

Do you have a link to the upstream bug report?

fleimgruber:
I neglected it for this long as well :)
FYI, I put up a bug report today, see http://lists.gnu.org/archive/html/bug-gnu-emacs/2015-02/msg00227.html.
I still need to provide tests for this, i.e. to show that `compile' fails in this specific case.

jorgenschaefer:
… ah, the supportive and helpful tone of emacs core development, it catches me off-guard every time :-D

In case it helps:

```Lisp
(defun compile-quoted (command &rest args)
  "Run `compile' with COMMAND and ARGS quoted.                                  
                                                                                
Useful if you can not be sure if there are special characters in                
the command or arguments, like `start-process' allows."
  (mapconcat #'shell-quote-argument
             (cons command args)
             " "))

(compile-quoted "write" "d:/asdf.txt")
(compile-quoted "write.test" "d:/asdf.txt")
```

Thanks for passing this on upstream.

fleimgruber:
Thanks for the use case sketch, I adapted it and put it up on the bug report.

fleimgruber:
FYI, this has been resolved upstream :)
http://lists.gnu.org/archive/html/bug-gnu-emacs/2015-02/msg00420.html

jorgenschaefer:
Thank you!

PR #428, opened 2014-12-28
Add support for Twisted's Trial test runner.

whirm:


jorgenschaefer:
Thank you for the patch! And sorry for the long delay, I was on vacation over the holidays and new year :-)

PR #451, opened 2015-01-23
Make tests less verbose by restricting yasnippet output.

birkenfeld:
I assume this is the right spot to put this?


jorgenschaefer:
Good idea! I'm unsure about `yas-snippet-dirs`, though – wouldn't that have the chance of possibly breaking some tests, or worse, possibly masking some test problems?

birkenfeld:
Without this, I did spot warnings about yasnippet checking for snippets in my ~/.emacs.d which should not be used I think.

Since the module-yasnippet test still passes, elpy's snippet addition is tested ok.

jorgenschaefer:
Guess we can fix the problems when they appear, instead of fixing imaginary problems! :-)

Thanks for the patch.

PR #463, opened 2015-01-29
Suggest running `elpy-config` after installation.

offby1:


jorgenschaefer:
Thank you very much for the contribution!

PR #474, opened 2015-02-04
Implement get_docstring for the Jedi backend.

birkenfeld:


jorgenschaefer:
… now I feel dumb for not having thought of that. :-D Thank you!

PR #484, opened 2015-02-08
Docstring: show what exact object the docstring is for

birkenfeld:
Reasoning: the Jedi docstring usually only includes the function name without full qualification. (And for modules, not even that.)

If you're fine with this change, I'll adapt the tests.


jorgenschaefer:
Looks good to me.

Please use `.format()` not `%` to stay consistent with the rest of Elpy, and try to avoid the backslash (parens allow multi-line expressions without the need for a backslash).

jorgenschaefer:
Thank you!

PR #485, opened 2015-02-08
Use popwin for docstring display if available.

birkenfeld:
Popwin is quite nice for such ephemeral windows. I don't know if that style of optional dependencies is ok with you though.

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -2052,6 +2052,12 @@ prefix argument is given, prompt for a symbol from the user."
           (symbol-name symbol)
         nil))))
 
+(when (require 'popwin nil t)
+  ;; display help buffer in a popup window using popwin
+  (add-to-list 'display-buffer-alist
+               '("*Python Doc*" popwin:special-display-popup-window)))
+
+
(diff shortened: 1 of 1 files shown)


jorgenschaefer:
Good idea, I guess. What does popwin do special? :-)

Please make this an Elpy module (using the `global-init` and `global-stop` methods to set this up), so users can disable this behavior if they dislike it.

birkenfeld:
`popwin` displays the buffer nicely in an unobtrusive way, and closes it automatically when it loses focus:

snap2

PS: The coveralls messages are getting a bit out of hand ;)

jorgenschaefer:
Nice module.

I wonder why coveralls is going crazy. :-D Wonder if it's configurable.

jorgenschaefer:
Note to self: This should be an elpy-module, not just global code.

gopar:
@birkenfeld any interest in making this into a elpy-module ?
If not then will have to close. Thanks

birkenfeld:
Sorry, no time ATM.

PR #492, opened 2015-02-11
Correctly use argument to elpy-use-ipython.

roshanshariff:


jorgenschaefer:
Oops, thank you for catching this and providing a fix! :-)

PR #506, opened 2015-02-22
Don't message "nil" when there is no error.

simenheg:
Calling `elpy-flymake-next-error`, `elpy-flymake-previous-error`, or `elpy-flymake-show-error` when there is no errors in the buffer, produces the message "nil", which isn't very helpful. This patch silences it. Or maybe displaying a proper message would be better?


jorgenschaefer:
Good catch, thank you for the patch!

PR #509, opened 2015-03-01
remove link to nonexistent documentation

ajschumacher:
there is nothing at https://github.com/jorgenschaefer/elpy/wiki/Keybindings
and I don't see any other page with keybinding info; probably better to not
have a link than to have a broken link.


jorgenschaefer:
Oops, forgot that reference. Thanks!

PR #529, opened 2015-04-03
recenter buffer when jumped to location

### LINKED ISSUE #528: elpy goto definition efficiently?
if i do `M-.` on any definition it takes me to the location of the file like this 

elpy1
(notice that the `point` is in the bottom line of buffer.)

but usually we jump to definition to take a look at it. so we have press `C-l` twice to take it to the top of buffer. 

so it makes sense to have a direct jump like this

elpy2

(notice that the `point` is in top line of buffer.)

ChillarAnand:
resolves #528

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -1558,7 +1558,8 @@ with a prefix argument)."
   (let ((buffer (find-file filename)))
     (with-current-buffer buffer
       (with-selected-window (get-buffer-window buffer)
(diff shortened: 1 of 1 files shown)


jorgenschaefer:
Thank you for the pull request! I implemented this in 007225c – using `recenter`, which is the correct underlying function that `recenter-top-bottom` calls if there is an argument.

PR #536, opened 2015-04-16
Fix docstring in elpy-test-*command variables

pallix:


jorgenschaefer:
Haha, good catch. Thank you! :-)

PR #542, opened 2015-04-20
Fix the interactive spec of elpy-nav-{for,back}ward-indent

cpitclaudel:
Adding "^" to the interactive spec of elpy-nav-forward-indent and
elpy-nav-backward-indent allows <C-left> and <C-right> to be used in
conjunction with shift to select blocks of text.


jorgenschaefer:
Nice! I did not know about that interactive spec, thank you for the code :-)

cpitclaudel:
My pleasure :)

PR #562, opened 2015-05-18
switch between python shell and buffer

### LINKED ISSUE #555: Switch back to python script buffer from python shell
First off, thank you for Elpy! It's a great package and you should be proud of it. I especially appreciate how responsive you are to your users. 

One thing I've wanted is the ability to switch back to the python script buffer from the inferior python shell. If you can switch to the shell with "C-c C-z", it would be really nice to have that same functionality to easily switch back after playing in the REPL for a bit. Oftentimes I have multiple windows open, and it's distracting to have to navigate back by switching with `other-window`. 

A great model for this feature is ESS's `ess-switch-to-inferior-or-script-buffer` command. It's smart enough to know which buffer you switched to shell from and will always take you back to it.

Thanks again for all your hard work.

ChillarAnand:
closes #555

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -310,7 +310,6 @@ edited instead. Setting this variable to nil disables this feature."
     ;; (define-key map (kbd "C-c <")   'python-indent-shift-left)
     ;; (define-key map (kbd "C-c >")   'python-indent-shift-right)
     (define-key map (kbd "C-c C-c") 'elpy-shell-send-region-or-buffer)
-    (define-key map (kbd "C-c C-z") 'elpy-shell-switch-to-shell)
     (define-key map (kbd "C-c C-d") 'elpy-doc)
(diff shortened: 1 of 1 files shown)


jorgenschaefer:
Hello, and thanks for the patch!
I do not think that Elpy's key map definition is the right place for this definition. Could you move it to `elpy-enable`? (Ideally, it'd be its own `elpy-module`, but that might be a bit overkill)

Also, could you please capitalize the commit  message and make sure it's understandable even a few years from now? Possibly add some added explanation to the body. Something like:

> Bind C-c C-z in Python shells to switch back to the Python buffer

http://chris.beams.io/posts/git-commit/ has a great summary on good commit messages :-)

ChillarAnand:
thanks for the article :)

jorgenschaefer:
Very nice! Thank you very much for another contribution. :-)

Issue #552, opened 2015-05-03
M-left/right matches region badly


jorgenschaefer:
```Python
1 if a:
2     if b:
3         if c:
4             pass
5         pass
6     pass
7 pass
```

Mark lines 1 – 5, hit M-right => the pass in line 5 is not moved
Hit M-left => the pass in line 6 is moved to the left

PR #567, opened 2015-05-21
Include python-indent when moving region left/right

ChillarAnand:
While moving a block left/right, empty space equivalent to
python-indent will be removed/added. So, it should be considered
while calculating effective end point of region.

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -1689,9 +1689,10 @@ indentation levels."
           (end (region-end)))
       (goto-char beg)
       (goto-char (point-at-bol))
-      (while (< (point) end)
+      (while (<= (point) end)
         (elpy--nav-move-line-left)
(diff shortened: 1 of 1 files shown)


ChillarAnand:
closes #552

jorgenschaefer:
Hm. I'm confused about this solution.

Using the example in #552:

```
1 if a:
2     if b:
3         if c:
4             pass
5         pass
6     pass
7 pass
```

If you mark lines 1 through 5, the line point is on is *not* highlighted using transient mark mode, so it should not move on `M-right`/`M-left`.

I guess for #552 to be fixed properly, we should come up with test cases first to agree on the correct behavior. Hm.

ChillarAnand:
Closing as this is no longer required.

PR #571, opened 2015-05-23
Automatically format code to PEP8 style using autopep8

### LINKED ISSUE #566: Add support for autopep8
As [mentioned here](https://github.com/jorgenschaefer/elpy/issues/565#issuecomment-103843256), it is a good idea to add autopep8 support for elpy.

ChillarAnand:
closes #566


jorgenschaefer:
First of all, thank you so much for this code! This is awesome.

I think it would be great if the autopep8 formatter would work on a region if one is selected, not unconditionally on the whole buffer.

Also, I think it would be nice to group the "code cleanup" stuff (`elpy-importmagic-fixup` and this) into a common key prefix to avoid cluttering too many key bindings, e.g. when we add the `isort` library. What do you think? (This could share the `C-c C-r` refactoring stuff …)

Lastly, out of curiosity, why did you remove `-Qwarnall` from the test runner?

ChillarAnand:
i saw this message
>  "Refactoring has been unstable and flakey, support will be dropped in the future."

 and thought, may be you are going to depricate this.

> Lastly, out of curiosity, why did you remove -Qwarnall from the test runner?

el

jorgenschaefer:
Yes, the current `C-c C-r` interface should go away at some point, and be replaced with something more sane.

Are you using Python 2 as `python`? That's a bit weird, but I guess the correct solution would be to use `python2` there to make it explicit.

ChillarAnand:
I am moving auto pep8  to elpy refactor. Check  out the code  & let me know if i am doing it right.  

> Lastly, out of curiosity, why did you remove -Qwarnall from the test runner?
-Q option was  removed  in Python3?
```
$ python3 -Q
Unknown option: -Q
usage: python3 [option] ... [-c cmd | -m mod | file | -] [arg] ...
Try `python -h' for more information.
(py3)
# anand at anand-X550EA in ~/projects/lisp/elpy on git:autopep8 x [23:21:41]
$ python2 -Q
Argument expected for the -Q option
```

jorgenschaefer:
I think there is a major misunderstanding here. I do not think adding autopep8 to the current refactor interface is worthwhile.

- Adding tons of special-purpose commands to do refactoring is a bad idea, so I would prefer not to do that. (Your first pull request)
- The current refactoring interface is terrible and needs to be replaced. No need to add anything to this.
- What I said was that we probably should figure out a better interface in the mid term.
- Until then, we probably can use special-purpose commands (as in your first proposal), but only under a single prefix key, not taking up a `C-c C-X` keys each.

Does that make it more clear what I mean? :-)

ChillarAnand:
So, we can have a function `elpy-code-cleanup` which does
importmagic-fixup,  autopep8 & isort and bind that function to a single
key(say C-c C-l)?



Regards,

Anand Reddy Pandikunta
www.avilpage.com

ChillarAnand:
>  I think it would be great if the autopep8 formatter would work on a region if one is selected, not unconditionally on the whole buffer.

That is a good idea. I have updated autopep8 to support that. 

> Also, I think it would be nice to group the "code cleanup" stuff (elpy-importmagic-fixup and this) into a common key prefix

Grouping these into a single key will be good if they are acting on whole buffer. However running impmagic fixup on region doesn't make sense as it is updating/sorting imports which are not in region.

jorgenschaefer:
> Grouping these into a single key will be good if they are acting on whole buffer. However running impmagic fixup on region doesn't make sense as it is updating/sorting imports which are not in region.

I think you misunderstood what I said – I was not talking about a single key, but about a _prefix_ key. E.g. have `C-c C-FOO i` call importmagic stuff, and `C-c C-FOO f` use autopep8. My biggest problem here is that I have no idea what key to use for `C-FOO`. Ideally, we'd eventually use `C-c C-r` for a generalized refactoring interface that does all of this in a sane way, but I do not see that any time soon. So we'd need an intermediate solution.

Do you have ideas?

whirm:
what about using hydra?

jorgenschaefer:
At least hydra-like would be an option, but that still needs an initial key sequence :-)

whirm:
I mean replacing the current refactoring menu with hydra, and keep using C-c C-r: if you have a region active, you can show different options than if you don't. And you tie all the refactoring stuff there, even the existing functionality. 

That could be done in a single step with not much effort, right? Or is there something I'm not getting?

jorgenschaefer:
No, that's quite fine. Except the current refactoring menu needs to be reworked.

(And I'm reluctant to add a dependency on hydra just for `C-c C-r`, but a hydra-like interface is pretty much what I had in mind for when I have a lot of time to work on this :-D)

whirm:
@jorgenschaefer what about magit-popup? Who doesn't have magit installed anyway? :D

jorgenschaefer:
> Who doesn't have magit installed anyway? :D

Me. :-)

ChillarAnand:
@jorgenschaefer since elpy is CLeaning the code, `C-c C-l` seems to be okay as prefix. 

 [makey](https://github.com/mickeynp/makey) looks like light weight solution for key binding interface?

personally, i use [guide-key](https://github.com/kai2nenobu/guide-key). when i press `C-c C-l`
aut

jorgenschaefer:
I do not think we need an external package for this feature until we flesh it out fully with more refactoring options (and even then we can easily do our own, this is a trivial use case).

`C-c C-l` is already taken. Although I do not know how well used it is, I guess it would be a bad idea to use a different key binding as an intermediate if we already know we're going to do something else.

So I think what I'd like here would be re-using `C-c C-r`, but keeping the old interface in a sub-key for that. Something like:

`C-c C-r i` / `C-c C-r C-i` for importmagic
`C-c C-r TAB` / `C-c C-r C-TAB` for autopep8 – alternatively, `C-c C-r 8` or `C-c C-r p` for *p*ep8? Preferences?
`C-c C-r r` / `C-c C-r C-r` for the current refactoring interface

Bonus points: Make `C-c C-r` display the selection options in the minibuffer.

Opinions?

whirm:
Looks fine to me.

ChillarAnand:
```
C-c C-r i - importmagic
C-c C-r r - refactoring
C-c C-r p - pep8ify
```
looks good to me :)

ChillarAnand:
Is this ok for showing [refactoring options](https://github.com/jorgenschaefer/elpy/commit/8e292a2fbd6debe41ca910bc55139d01a851e530)?

jorgenschaefer:
I like that!

jorgenschaefer:
As a note, I haven't forgotten your contributions, I'm just quite busy as of late. I will have more time to work on Elpy mid-August I hope. Sorry about that!

ChillarAnand:
No issues :-) 
Yesterday I was reading about `interactive c` and thought may be that is sufficient for refactoring options.

jorgenschaefer:
Ok, looking over the code, I definitely like this. Could you rebase the PR onto of the master branch, and squash it to a single commit? Thank you!

abdulhaq-e:
Is there a way to turn this off without uninstalling autopep8?

ChillarAnand:
To format code as per PEP8 conventions, you need to run `C-c C-r p` which will do formatting.

Elpy will not automatically format code as you type.

PR #583, opened 2015-06-05
update the documentation to match last block indent changes

Chedi:
just an update in relation with the commit fb68aa01bb4ccce0420766264e8757c2b2487ac8 about the block indent removal


jorgenschaefer:
Good catch, thank you very much for the contribution!

PR #587, opened 2015-06-10
Use custom symbol for regexp search

### LINKED ISSUE #574: Include html files while grepping?
While working on django projects, if user searches for variable `foo` only *.py files are grepped. 

For this we can use projectile-grep but it does fuzzy match and it is very slower.

Also if we want to rename a variable in a view and use multi edit, it wont edit variables in template files.

May be we can include *.html files?

ChillarAnand:
closes #574

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -295,6 +295,11 @@ edited instead. Setting this variable to nil disables this feature."
   :type '(repeat string)
   :group 'elpy)
 
+(defcustom elpy-rgrep-file-pattern "*.py"
+  "FILES to use for `elpy-rgrep-symbol'."
+  :type 'string
+  :group 'elpy)
+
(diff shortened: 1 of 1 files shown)


jorgenschaefer:
Thanks for the contribution!

`elpy-rgrep-symbol` is a confusing name – what do you think about `elpy-rgrep-file-pattern` instead?
Also, could you be a bit more verbose in the commit message, explaining what this change is doing?

ChillarAnand:
elpy-importmagic-fixup-test is failing on emacs 24.1. Dont know why.

jorgenschaefer:
Hm. Temporary failure, apparently. That's weird. Flaky tests are annoying.

jorgenschaefer:
Thanks again for the contribution!

PR #617, opened 2015-07-22
Add function to select surrounding lines with current indentation

### LINKED ISSUE #616: Select surrounding lines with same indentation
As a user, I want to be able to select all surrounding lines with the same indentation as the current one, in order to manipulate this region (e.g. changing indentation).

Sublime apparently calls this "expand to indentation" and binds it to `C-S-j`

ChillarAnand:
Not sure which key binding to use for this. Suggestions?

closes #616

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -309,6 +309,7 @@ edited instead. Setting this variable to nil disables this feature."
     ;; (define-key map (kbd "C-M-x")   'python-shell-send-defun)
     ;; (define-key map (kbd "C-c <")   'python-indent-shift-left)
     ;; (define-key map (kbd "C-c >")   'python-indent-shift-right)
+    (define-key map (kbd "C-c C-b") 'elpy-nav-expand-to-indentation)
     (define-key map (kbd "C-c C-c") 'elpy-shell-send-region-or-buffer)
     (define-key map (kbd "C-c C-d") 'elpy-doc)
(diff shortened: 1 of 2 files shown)


jorgenschaefer:
Thank you for this work!

I think #616 actually means "same or deeper indentation", so you shouldn't use `eq` but rather something like `<=` for your checks (in general, don't use `eq` for numbers, just use `=` or at least `eql`; there is no guarantee that two identical numbers are `eq` if Emacs ever gets bignums, like XEmacs has).

Also please note that `previous-line` and `next-line` say specifically _"This function is for interactive use only"_, so please use `forward-line` as advised there. The correct way of marking and highlighting a region from Emacs Lisp is to use `set-mark` and `activate-mark`, or `push-mark` with the `ACTIVATE` argument set to `t`, so something like this:

```Lisp
(goto-char end)
(push-mark beg nil t)
```

You can try writing a test for this, too, if you like, but that can be tricky :-)

ChillarAnand:
My bad. Lots of mistakes!

Updated function & added a test case.

jorgenschaefer:
Wow, very nice! This looks good.

The key binding `C-c C-j` is used by python.el already. What do you think about `C-c C-b` for "block"?

Other than that, I'm happy to merge this – could you squash the commits into a single one?

Thank you!

ChillarAnand:
Done!

jorgenschaefer:
Excellent. Thank you very much for your work! :-)

PR #619, opened 2015-07-26
Add function to send current statement to Python shell

### LINKED ISSUE #381: C-RET should send current statement
This is what ESS does, and which can be useful in Python, too, especially in combination with ipython.

ChillarAnand:
@jorgenschaefer  I tried  to add a test case similar to `elpy-shell-send-region-or-buffer-should-send-region-if-active`, but it is executing entire buffer?

closes #381

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -328,6 +328,8 @@ edited instead. Setting this variable to nil disables this feature."
     (define-key map (kbd "<S-return>") 'elpy-open-and-indent-line-below)
     (define-key map (kbd "<C-S-return>") 'elpy-open-and-indent-line-above)
 
+    (define-key map (kbd "<C-return>") 'elpy-shell-send-current-statement)
+
     (define-key map (kbd "<C-down>") 'elpy-nav-forward-block)
     (define-key map (kbd "<C-up>") 'elpy-nav-backward-block)
     (define-key map (kbd "<C-left>") 'elpy-nav-backward-indent)
@@ -1416,6 +1418,12 @@ else:
    (t
     (error "I don't know how to set ipython settings for this Emacs"))))
 
+(defun elpy-shell-display-buffer ()
+  "Display inferior Python process buffer."
+  (display-buffer (process-buffer (elpy-shell-get-or-create-process))
+                  nil
+                  'visible))
+
 (defun elpy-shell-send-region-or-buffer (&optional arg)
   "Send the active region or the buffer to the Python shell.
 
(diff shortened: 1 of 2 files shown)


ChillarAnand:
Test has passed :D

jorgenschaefer:
Thank you for the work!

Why are you marking the region you are sending? Do you think the user expects `<C-return>` to interfere with their mark? In general, from Lisp, you should not change point and mark, unless that is specifically what you want to happen (like in your last patch, creating an active region was the goal of the code). You can read more about this in the docstring of `set-mark`. Usually, Lisp code will calculate the right positions and just store them in variables, to send them as needed.

Incidentally, do you think the user of this feature would like point to end up at the beginning of the next statement, so that successive calls to `C-return` would send a statement at a time?

ChillarAnand:
I thought we can re use code by setting mark and using `elpy send region or buffer` function. 

Yes, elpy shouldnt change mark for this.

ChillarAnand:
Also when running a single test `cask exec ert-runner test/elpy-shell-send-current-statement.el`, it is failing but when running entire test suite, test is passing?

jorgenschaefer:
> I thought we can re use code by setting mark and using `elpy send region or buffer` function.

Reusing code is a good idea, but in this case, the right solution would be to extract common code into a helper function, not to fake user input :-)

> Updated now.

Thank you! Any reason not to use `python-shell-send-region` (and pass `SEND-MAIN` as `t` in this case, I guess)?

> Also when running a single test `cask exec ert-runner test/elpy-shell-send-current-statement.el`, it is failing but when running entire test suite, test is passing?

You're not enabling python-mode/elpy-mode in the buffer, I think?

Oh, `elpy-testcase` should indent the body only 2 spaces. If you `load-file` the `test-helper.el` from `elpy/test/`, Emacs should do the right thing.

ChillarAnand:
> Any reason not to use python-shell-send-region?

It is throwing indentation error.

jorgenschaefer:
>> Any reason not to use python-shell-send-region?
> It is throwing indentation error.

Hm. Strange. Do you have a reproduction case for that?

ChillarAnand:
Any indented statement is throwing error.

    if True:
         pr_|_int('a')

Code:

    (let ((beg (python-nav-beginning-of-statement))
          (end (python-nav-end-of-statement)))
      (elpy-shell-get-or-create-process)
      (python-shell-send-region beg end))

jorgenschaefer:
Fancy. That's actually a bug in `python-shell-buffer-substring`. :-D Reported upstream as [debbugs#21193](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=21193).

jorgenschaefer:
Very nice. Again, thank you for the contribution!

PR #625, opened 2015-08-08
Add `elpy-test-do-at-point-p`

### LINKED ISSUE #624: elpy-test-at-point is too restrictive about test file names
This condition is used to determine whether elpy-test-at-point should return test information

``` elisp
(if (and file (string-match "/test[^/]*$" file))
```

Some python libraries do not use this naming convention for test files. Is this logic even all that necessary. It seems like it should be enough if the function in close proximity has a 'test' name.

If you think that keeping this is absolutely necessary I would at least recommend that it be moved into a function that is configurable by the user.

something like `elpy-test--is-test-file-function`

I can make this change if it sounds good to you.

IvanMalison:
This variable will control when at point information should
be used for running individual tests. Fixes #624.

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -269,6 +269,13 @@ edited instead. Setting this variable to nil disables this feature."
   :safe 'elpy-test-runner-p
   :group 'elpy)
 
+(defcustom elpy-test-do-at-point-p 'elpy-any-has-test-p
+  "Predicate that decides whether the module/class/test name
+information should be used to run a specific test or not"
+  :type '(choice (const :tag "Any of module/class/function name has test" elpy-any-has-test-p)
+                 (const :tag "All of module/class/function name have test" elpy-all-have-test-p))
+  :group 'elpy)
+
 (defcustom elpy-test-discover-runner-command '("python" "-m" "unittest")
   "The command to use for `elpy-test-discover-runner'."
   :type '(repeat string)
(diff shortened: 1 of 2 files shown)


IvanMalison:
I did what I suggested in #624 but after I thought about it a bit, I decided that it might make sense to restrict when this happens or how exactly it happens. What happens if you are in a setup method for a test class. With this pull request, the runner would attempt to run the test class . the setup method which would likely result in no tests being executed. Perhaps there should be a transformation step on the test name? I guess the individual runners could handle this? It's hard to support all these different test runners in a generic way :(

jorgenschaefer:

When do you think a user would seriously change the value of that option from the default? Considering the default heuristic is easily circumvented – use a prefix argument or just go to the start/end of the buffer before hitting `C-c C-t` – the extra option seems a bit superfluous.

Do you yourself have a need for such an option?

IvanMalison:
> When do you think a user would seriously change the value of that option from the default?

I doubt they would. I only added it in case someone preferred the old behavior.

IvanMalison:
Looks like the tests failed, which is strange because I ran them locally and they passed. Maybe it has to do with version?

IvanMalison:
ah yeah looks like they fail for 24.1 and 24.2. Interesting. Will investigate.

jorgenschaefer:
> I doubt they would. I only added it in case someone preferred the old behavior.

Leave it out, then. I like the new behavior you suggested. :-)

jorgenschaefer:
Any updates on this code? The tests are failing, and I think the simplification we talked about above would be nice :-)

IvanMalison:
I'll try to look this weekend

jorgenschaefer:
Hello! Any updates on this?

gopar:
Update?

gopar:
Over 2 years with no updates. Closing this

PR #641, opened 2015-08-25
Select whole buffer for zero indent expansion

ChillarAnand:


jorgenschaefer:
Thanks for the contribution. Could you explain (ideally in the commit message) why you are doing this and what use case this addresses? Also, it's a good idea to mention the context in the commit message, for example by mentioning the relevant function name.

ChillarAnand:
Right now, when point is in line with zero indentation, calling `elpy-nav-expand-to-indenation` makes elpy to 'hang'. This patch fixes it.

jorgenschaefer:
Ah, that makes sense. Could you write this in the commit message? Ideally, the title line should be a short summary of what is being done (like, "expand-to-indentation: Mark whole buffer without current indentation"), and then an explanation as to _why_ this change was done. Remember, the changelog is there so you can figure out what happened a year ago there, and why you did this. :-)

ChillarAnand:
Merge this?

jorgenschaefer:
Ah, apparently github does not send notifications when a PR is updated, only when a comment is written. Thanks for the changes and this improvement!

PR #651, opened 2015-08-29
Delete instead of deindent when code is selected up to the indentation

Sbte:
If I select abc from right to left in this piece of code

```
        abcde
```
and then hit backspace, this would result in
```
    abcde
```
instead I'd expect it to be
```
        de
```
which is the same you'd get when selecting from left to right. This pull request attempts to fix this. Reason I use backspace instead of delete is because the delete key appears to be located at a semi-random place on different keyboards. Especially laptop keyboards.

Since this is my first piece of lisp code ever, please check it very carefully and any advice on how to improve it would be welcome. The code is based on the corresponding function in python.el

https://github.com/fgallina/python.el/blob/master/python.el#L973

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -308,7 +308,8 @@ edited instead. Setting this variable to nil disables this feature."
     ;; Alphabetical order to make it easier to find free C-c C-X
     ;; bindings in the future. Heh.
 
-    ;; (define-key map (kbd "<backspace>") 'python-indent-dedent-line-backspace)
+
+    (define-key map (kbd "<backspace>") 'elpy-indent-dedent-line-backspace)
     ;; (define-key map (kbd "<backtab>")   'python-indent-dedent-line)
 
     ;; (define-key map (kbd "C-M-x")   'python-shell-send-defun)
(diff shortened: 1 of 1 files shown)


jorgenschaefer:
Thank you for the code! This is indeed what I'd expect, too. Sadly, I can not reproduce the behavior you describe there. Which version of Emacs are you using?

Sbte:
Version 24.3.1. Since this code is still in the python.el master branch, I don't know what could have fixed it? Did you select from right to left?

jorgenschaefer:
> Version 24.3.1. Since this code is still in the python.el master branch, I don't know what could have fixed it?

Hm. I can not reproduce this with 24.3, either.

> Did you select from right to left?

Yes, I have.

I started Emacs with `emacs-24.3 -Q`, enabled `delete-selection-mode`, opened a new Python file, entered 8 spaces, then `abcde`, hit `C-SPC on `d`, moved to the left on `a`, and hit backspace. I ended up with a file containing 8 spaces, then `de`.

Can you provide reproduction steps for an `emacs -Q` case?

Sbte:
Ok, managed to reproduce it.

`emacs -Q test.py`

type

```
def f():
    abcde

```
Make sure you have the newline. Not hit left arrow 3 times, hold shift, and hit the left arrow another 3 times. Now hit backspace (sometimes 1, sometimes 2 times, not sure why). The line will now be dedented. abc will not be deleted.

jorgenschaefer:
Aha! Thank you very much, this does indeed reproduce it for me. :-)

jorgenschaefer:
Ok, sorry for the delay here. As this is a bug in python.el, I suspect this should be fixed there. Does the following fix the problem for you?

```Lisp
(defun python-indent-dedent-line-backspace (arg)
  "De-indent current line.                                                      
Argument ARG is passed to `backward-delete-char-untabify' when                  
point is not in between the indentation."
  (interactive "*p")
  (when (or (region-active-p)
            (not (python-indent-dedent-line)))
    (backward-delete-char-untabify arg)))
```

jorgenschaefer:
I assume this problem is resolved. Than you for helping making Elpy better, and do not hesitate to open a new issue or PR if you have further problems! :-)

PR #659, opened 2015-09-04
Ensure rope is installed before refactoring code

### LINKED ISSUE #658: elpy-refactor should give an error message if tried without rope installed
Right now it just returns an empty buffer which does not seem that user friendly:)

ChillarAnand:
closes #658


jorgenschaefer:
Hm. Apparently, server.py already had some checks for import errors which somehow got left out a while ago … thanks.

Issue #661, opened 2015-09-08
elpy-nav-indent-shift-right does shift unmarked last line


jorgenschaefer:
With mark at `_!_` and point at `_|_`, I expect `elpy-nav-indent-shift-right` not to indent the line `baz`:

```
_!_foo
bar
_|_baz
```

Currently, it does.

jorgenschaefer:
Fixed in 1750024

PR #662, opened 2015-09-08
Retain point while autopep8 refactoring

### LINKED ISSUE #660: Could autopep8 leave the cursor where it was?
At the moment elpy-autopep8-fix-code moves the cursor to the beginning of the file after the cleanup is done. Is it possible to change this behaviour to leave the cursor as close as possible to where it was when the function was called?

ChillarAnand:
closes #660 #661

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -3426,13 +3426,17 @@ description."
 (defun elpy-autopep8-fix-code ()
   "Automatically formats Python code to conform to the PEP 8 style guide."
   (interactive)
-  (if (use-region-p)
-    (let ((new-block (elpy-rpc "fix_code" (list (elpy-rpc--region-contents))))
-          (beg (region-beginning)) (end (region-end)))
-      (elpy-buffer--replace-region beg end new-block))
-    (let ((new-block (elpy-rpc "fix_code" (list (elpy-rpc--buffer-contents))))
-          (beg (point-min)) (end (point-max)))
-      (elpy-buffer--replace-region beg end new-block))))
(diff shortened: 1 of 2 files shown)


jorgenschaefer:
- Hm. Retaining point after replacing the whole buffer is tricky. Maybe rather store line number and column or something?
- Please use `when` instead of `if` if you have no alternative branch. :-)

ChillarAnand:
> Hm. Retaining point after replacing the whole buffer is tricky. Maybe rather store line number and column or something?

autopep8 does whitespace cleanup, there is no gaurantee that saved line/column will even exist after cleanup.

> Please use when instead of if if you have no alternative branch. :-)

Roger that :-)

jorgenschaefer:
Applied the partially selected line fix. Thank you, excellent work! I'm especially happy for the tests! :-)

> autopep8 does whitespace cleanup, there is no gaurantee that saved line/column will even exist after cleanup.

Nor is there a guarantee that the saved byte offset into the buffer will exist, which is what your patch is adding. Both are heuristics and guesswork. The question is, which one will work better in most cases?

Using the byte offset means that even small changes will cause point to move around. Line and column might be less confusing for users.

A user on `#emacs-elpy` mentioned that `py-autopep8.el` is able to retain cursor position more accurately than Elpy. Any idea how it does this?

ChillarAnand:
Yes, yes. It is much better to restore line/column positions. py-autopep8 is also doing same thing. Updated it. :-)

jorgenschaefer:
Excellent, thank you!

PR #667, opened 2015-09-16
Restart rpc process when virtualenv is activated

### LINKED ISSUE #663: Restart rpc after changing venv
It is better if elpy can restart rpc automatically after changing a virtualenv.

ChillarAnand:
closes #663

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -431,7 +431,8 @@ edited instead. Setting this variable to nil disables this feature."
                      "Emacs 24 and above"))))
   (elpy-modules-global-init)
   (define-key inferior-python-mode-map (kbd "C-c C-z") 'elpy-shell-switch-to-buffer)
-  (add-hook 'python-mode-hook 'elpy-mode))
+  (add-hook 'python-mode-hook 'elpy-mode)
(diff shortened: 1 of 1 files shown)


jorgenschaefer:
Thanks :-) `elpy-rpc--buffer` is a buffer-local variable, and `pyvenv-post-activate-hooks` is not run in any specific buffer, so you will need to find Elpy buffers to reset that variable in them.

Also, the `add-hook` call should be in `elpy-enable`, not at the top level.

ChillarAnand:
Wouldn't it be  easier if directly call `elpy-rpc-restart`?

jorgenschaefer:
> Wouldn't it be easier if directly call `elpy-rpc-restart`?

Depends. `elpy-rpc-restart` would kill all existing RPC processes. Do we want to keep the old RPC process around in case we switch back to the old virtualenv (e.g. using pyvenv's tracking mode), or do we want to re-start them every time, possibly causing delays with importmagic's loader?

ChillarAnand:
I didn't know about that. Updated :-)

jorgenschaefer:
Excellent, thank you!

PR #672, opened 2015-09-21
Retain point when expanding to indentation

ChillarAnand:
After executing `expand-to-indentation`, if `C-g` is pressed, point
should return to original position

Test is failing, don't know how to  handle `keyboard-quit`.

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -431,7 +431,8 @@ edited instead. Setting this variable to nil disables this feature."
                      "Emacs 24 and above"))))
   (elpy-modules-global-init)
   (define-key inferior-python-mode-map (kbd "C-c C-z") 'elpy-shell-switch-to-buffer)
-  (add-hook 'python-mode-hook 'elpy-mode))
+  (add-hook 'python-mode-hook 'elpy-mode)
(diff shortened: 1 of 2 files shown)


jorgenschaefer:
Hm. I can see the use case for this, but I feel very uncomfortable adding a post-command hook just for this. That seems rather excessive for such a minor nuisance. Is this really necessary?

And yes, Emacs does not make it easy to test keyboard interrupt effects. It's probably best to just call `elpy-nav--restore-initial-position` directly there.

ChillarAnand:
> Is this really necessary?

Don't know. Most packages like expand-region, web-mode e.t.c. implement this and i got used to that behaviour.

I think expand-region does this without hooks, if it is simple enough, we can try that.

jorgenschaefer:
Woah :-D https://github.com/magnars/expand-region.el/blob/master/expand-region-core.el#L242-L253

Yeah, better solution, but wow that's ugly. Wouldn't mind a similar implementation if you find that useful.

ChillarAnand:
I know only those solutions. I think it is better to leave elpy as it as for now.

PR #673, opened 2015-09-23
Retain point when expanding to indentation

ChillarAnand:
I find it useful, but i am still skeptical whether it should be added to elpy or not.


jorgenschaefer:
Much better, thanks a lot!

Could you change the `memq` call to `eq`, as you only have one item to compare to?

ChillarAnand:
Yes please :-)

jorgenschaefer:
Excellent! :-)

PR #683, opened 2015-10-08
Rope shouldn't refactor without project root

### LINKED ISSUE #682: rope -  refactoring error
Elpy Error

The backend encountered an unexpected error. This indicates a bug in
Elpy. Please open a bug report with the data below in the Elpy bug
tracker:

https://github.com/jorgenschaefer/elpy/issues/new

```
Error Message

[Errno 2] No such file or directory: '/ssh:foo@bar:/baz'

Configuration

Virtualenv........: exp (/home/k3/.virtualenvs/exp)
RPC Python........: 2.7.6 (/home/k3/.virtualenvs/exp/bin/python)
Interactive Python: python (/home/k3/.virtualenvs/exp/bin/python)
Emacs.............: 24.4.1
Elpy..............: 1.9.0
Jedi..............: 0.9.0
Rope..............: 0.10.2
Importmagic.......: 0.1.3
Autopep8..........: 0.1.3
Syntax checker....: pyflakes (/home/k3/.virtualenvs/exp/bin/pyflakes)

Traceback

Traceback (most recent call last):
  File "/home/k3/.emacs.d/elpa/elpy-20150830.510/elpy/rpc.py", line 100, in handle_request
    result = method(*params)
  File "/home/k3/.emacs.d/elpa/elpy-20150830.510/elpy/server.py", line 170, in rpc_get_refactor_options
    ref = refactor.Refactor(self.project_root, filename)
  File "/home/k3/.emacs.d/elpa/elpy-20150830.510/elpy/refactor.py", line 101, in __init__
    self.project = Project(project_root, ropefolder=None)
  File "/home/k3/.virtualenvs/exp/local/lib/python2.7/site-packages/rope/base/project.py", line 134, in __init__
    os.mkdir(self._address)
OSError: [Errno 2] No such file or directory: '/ssh:foo@bar:/baz'
```

I guess it is better if  rope doesn't do any refactor in tramp files.

ChillarAnand:
closes #682

### DIFF
--- a/elpy/refactor.py
+++ b/elpy/refactor.py
@@ -50,6 +50,8 @@ something wrong.
 
 """
 
+import os
+
 from elpy.rpc import Fault
 
 try:
@@ -97,12 +99,16 @@ class Refactor(object):
     """
     def __init__(self, project_root, filename):
         self.project_root = project_root
-        if ROPE_AVAILABLE:
-            self.project = Project(project_root, ropefolder=None)
-            self.resource = path_to_resource(self.project, filename)
-        else:
+        if not ROPE_AVAILABLE:
             raise Fault('rope not installed, cannot refactor code.',
                         code=400)
+        if not os.path.exists(project_root):
+            raise Fault(
+                "cannot do refactoring without a local project root",
+                code=400
+            )
+        self.project = Project(project_root, ropefolder=None)
+        self.resource = path_to_resource(self.project, filename)
 
     def get_refactor_options(self, start, end=None):
         """Return a list of options for refactoring at the given position.


jorgenschaefer:
Excellent commit message!

I think it is possible for `project_root` to be `None`, so maybe check for `not project_root or os.path.exists(project_root)`?

ChillarAnand:
if we check for `not project_root or os.path.exists(project_root)` and if `project_root` is `/ssh:foo@bar:/baz`,  then condition will fail?

is it ok to check for  `project_root and os.path.exists(project_root)`?

jorgenschaefer:
Sure. You know what I meant :-)

ChillarAnand:
I have refactored top if condition, will that be fine?

ChillarAnand:
Should we do check  for `None` here  too? https://github.com/jorgenschaefer/elpy/blob/master/elpy/ropebackend.py#L40

jorgenschaefer:
Hm. Good point. That hasn't crashed so far, so I suspect there's very rarely no project root. Come to think of it, I think Elpy will resort to the current directory for project root if everything else fails? So I guess I was wrong and the check for None is useless. Damnit. :-) Sorry about this.

ChillarAnand:
No  issues. Updated  :-)

jorgenschaefer:
That looks very good. I like doing the fault checks first. Thanks a lot!

ChillarAnand:
:+1:

PR #684, opened 2015-10-09
Indent outdented code

ChillarAnand:
I have been using couple of hooks which   indent  code  automatically. Whenever i open elpy.el these 2   blocks will get indented  properly and while  staging it becomes  a problem as  i  have to stage entire file  and then  unstage   these 2 blocks. So just submitting this.


jorgenschaefer:
Thanks!

PR #700, opened 2015-10-28
fixed elpy-goto-location: find-file returns nil on 24.5.1

lins05:
- OS: ubuntu 14.04 LTS 64bit
- Emacs Version: 24.5.1
- Elpy installed from latest source

Every time I invoke `elpy-goto-definition`, emacs gave me error `save-current-buffer: Wrong type argument: stringp, nil`. Using edebug I find it's because ` find-file` returns nil.

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -1610,7 +1610,8 @@ with a prefix argument)."
 (defun elpy-goto-location (filename offset)
   "Show FILENAME at OFFSET to the user."
   (ring-insert find-tag-marker-ring (point-marker))
-  (let ((buffer (find-file filename)))
+  (find-file filename)
+  (let ((buffer (get-file-buffer filename)))
     (with-current-buffer buffer
       (with-selected-window (get-buffer-window buffer)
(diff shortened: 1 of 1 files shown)


jorgenschaefer:
Hello, and thank you for the contribution!
I am a bit confused, though – [the manual on find-file](https://www.gnu.org/software/emacs/manual/html_node/elisp/Visiting-Functions.html) is rather explicit in saying that _"[i]t also returns that buffer."_ Are you maybe using a package that replaces `find-file` with an incompatible implementation or advice?

lins05:
@jorgenschaefer You are right, the return value of `find-file` is affected by advices added by another package I use. I've sent a [PR](https://github.com/cute-jumper/fcitx.el/pull/4) to that, so close this one now. 

Again, thanks for this wonderful package!

jorgenschaefer:
Great! Thank you for the kind words, and especially for the work you put into fixing this issue. Enjoy Elpy! :-)

PR #706, opened 2015-11-06
Typo: use correct keys to fetch values from config

ChillarAnand:


jorgenschaefer:
Oops! :-) Thanks for the fix!

PR #707, opened 2015-11-06
Include yapf in elpy config

ChillarAnand:
Added yapf to elpy config.
Key binding(`C-c C-r y`) to call yapf formatter.

### DIFF
--- a/docs/ide.rst
+++ b/docs/ide.rst
@@ -325,22 +325,24 @@ Elpy supports various forms of refactoring Python code.
    occurrences of the text in the region.
 
 
+.. command:: elpy-format-code
+   :kbd: C-c C-r f
+
+   Format code using the available formatter.
+
+   This command formats code using `yapf`_ or `autopep8`_ formatter. If a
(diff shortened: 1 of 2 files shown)


jorgenschaefer:
Thanks for the contribution!

I am currently wondering if it would be better maybe to have one key binding for "fix code formatting", and then use yapf or autopep8 depending on availability? What do you think?

ChillarAnand:
When both are available, which one should we use?

jorgenschaefer:
> When both are available, which one should we use?

I'd say yapf, as it is more featureful.

Do you think there is a (realistic) use case where someone has both autopep8 and yapf installed and would want to run autopep8 instead of yapf?

ChillarAnand:
Yes it is better to make `yapf` default formatter.

Sorry for the delay.

jorgenschaefer:
Thank you!

abdulhaq-e:
This is not included in the latest released version is it?

Thanks.

jorgenschaefer:
I'm afraid it is not, no. It will be in the next release.

PR #728, opened 2015-12-01
Add ^ specifier to two more -nav commands

cpitclaudel:
Looks like I left these two out in my last PR from a while back.


jorgenschaefer:
Thank you!

cpitclaudel:
:)

PR #741, opened 2015-12-14
Fixed a bug (on my end) removing incorrect args

KennethAdamMiller:
The run-python command was shorting out. I fixed this some time ago,
but never took the time to share it with the community. Here it is; it worked
on my end when I used it, but someone please do a checkout and test, because
I hate having "it runs on my box" issues, and I would be sorely pissed at myself
if I ruined anybody else's experience causing unnecessary complications. 
Maybe it's small, but hopefully it helps :D

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -1565,8 +1565,15 @@ code is executed."
          (proc (get-buffer-process bufname)))
     (if proc
         proc
-      (run-python (python-shell-parse-command))
-      (get-buffer-process bufname))))
+      (run-python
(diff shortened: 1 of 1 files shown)


KennethAdamMiller:
Let me just fix the build issue with travis and give it back afterward.

jorgenschaefer:
Hello, and thank you for the contribution. Sorry for not responding earlier, I was on vacation. Looking forward to the final version. :-)

Issue #643, opened 2015-08-25
Insert parens while completing functions


ChillarAnand:
Since we call functions most of the time, it is better to insert parens by default.

Ref: http://emacs.stackexchange.com/q/15032/5187

jorgenschaefer:
Thanks for the idea. I'm not sure this would be a good idea. When using electric-pair-mode (or similar packages), typing the opening parenthesis inserts both parens and puts the cursor on the closing paren. Company does not support this kind of behavior, and I would it incredibly annoying if I had to type `<left>` after completing a function all the time.

How exactly did you think this would be done?

ChillarAnand:
I am not familiar with company mode and have no idea how to implement it. :(

jorgenschaefer:
Ah!

Yeah, I think it would be nice if Elpy could complete a function with the parentheses at the end, but only if it could put point on the closing parenthesis after completion. Something to be figured out in how to support this with company. (And, considering people use various parenthesis tools, like smartparens, probably something that should be optional …)

ChillarAnand:
May be this will help http://emacs.stackexchange.com/a/15035/5187

jorgenschaefer:
Adding parentheses to the completion if it's a function isn't the problem, really (I think the backend can do it, or we just annotate it). The problem is not moving point after the closing parenthesis.

PR #756, opened 2016-01-05
Complete Python callables optionally with parentheses

gav451:
Resolves issue https://github.com/jorgenschaefer/elpy/issues/643, since
point is left between the parentheses.

In case that elpy-company-add-parentheses-after-callables is non-nil,
insert parentheses after completions:
- With class and function annotations.
- With instance and statement annotations in case that the completions
  have a calltip.


jorgenschaefer:
Hello, and thank you for the contribution!

The code should only insert anything at all if the completion is a callable, not insert parens and then delete them if there is not.

Sadly, even then this code will block on every invocation, slowing down completion and general interaction. Instead of using `elpy-rpc-get-calltip`, could you use `elpy-company--cache-annotation` and see if that returns `"function"`? That should be a lot faster and have the same effect.

gav451:
Initially, I had written code only to append parentheses when "elpy-company--cache-annotation" returned "function" or "class" to append parentheses.
Then I discovered that this solution did not cover for instance universal functions in numpy such as numpy.abs which has the following entry in the elpy-company

               ((annotation . "instance")
                (meta . "instance: numpy.core.abs")
                (name . "abs")
                (suffix . "bs"))

jorgenschaefer:
> Then I discovered that this solution did not cover for instance universal functions in numpy such as numpy.abs

Hm yeah. Numpy is notoriously difficult to handle correctly for jedi. Might be worth a bug report there!

I just realized that you used `post-completion` in your original code – that means it's not as bad for completion as I worried. I can live with that. If the annotation is a problem, could you modify your code to at least insert the parentheses only when needed?

gav451:
The code now allows the user to provide his own post completion function, because I think that adding 
parentheses after callables depends a lot on the Emacs and Python environment as well as the coding 
style. I think that my post completion functions are correct in 90 % of my use cases and they show that
it will be very hard to write such a function that "does always what I mean".

gav451:
Sorry for the confusion: I wanted to amend my commit, but I failed to do it properly.

jorgenschaefer:
Thank you. Looks like github does not notify me when a PR is updated.

I like this a lot.

Is there a drawback to just always use the numpy version? `instance` is the name for old-style classes.

Also, I think you can simplify the exception detection a lot by simply doing something like `(progn (backward-sexp 2) (looking-at "\\_<raise\\_>"))` – much simpler and more general. How does that sound?

gav451:
I do not think that the numpy version has any problems, since the only problem I can imagine it is that completion is too slow for interactive use, but in my experience that is not the case at all.
I will rename elpy-company-numpy-post-complete-parens to elpy-company-post-complete-parens to get rid of the non-numpy version.
I did not realize that checking for a raise statement could possibly be as simple as you suggest. Please, give until the week-end to do my home work ;-) 
I really appreciate how you push/guide/teach your contributors (and elpy consists of very clear code!).

gav451:
On 03/01/2016 08:02 PM, Jorgen Schäfer wrote:
>
> Thank you. Looks like github does not notify me when a PR is updated.
>
> I like this a lot.
>
> Is there a drawback to just always use the numpy version? |instance| 
> is the name for old-style classes.
>
> Also, I think you can simplify the exception detection a lot by simply 
> doing something like |(progn (backward-sexp 2) (looking-at 
> "\\_<raise\\_>"))| – much simpler and more general. How does that sound?
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/jorgenschaefer/elpy/pull/756#issuecomment-190853178>.
>
I updated the pull request taking into account your remarks.
However, I used (backward-word 2) instead of (backward-sexp 2) since the 
former accepts raise statements containing a line continuation character 
('\').

jorgenschaefer:
Looks good to me, thank you! Could you squash the commits in this PR into one and force push?

jorgenschaefer:
(If you want, some unit tests for this behavior, especially the "accepts raise statements with continuation character" thing would be great, but not required :-))

gav451:
On 03/06/2016 12:41 PM, Jorgen Schäfer wrote:
>
> Looks good to me, thank you! Could you squash the commits in this PR 
> into one and force push?
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/jorgenschaefer/elpy/pull/756#issuecomment-192878550>.
>
I managed to squash the commits in the PR and to force push them.
Providing a unit test is a next step in learning elpy.

jorgenschaefer:
Hm. There are still 11 commits in this pull request. Did you rebase your final commit on top of the master branch? :-)

gav451:
I think I have squashed all my commits in my 'parentheses-after-callables' branch, because looking at https://github.com/gav451/elpy/commits/parentheses-after-callables I see only one commit by me
with the original date of January 5th, but it contains a squash of all my later commits.
Maybe some confusion arises from pulling changes from your master into my 'parentheses-after-callables' branch, because that is why the other commits are there.
Do I have to push this single commit on my master before you can pull it, or can you pull it from my 'parentheses-after-callables' branch?

jorgenschaefer:
The branch indeed only contained one commit, but the branch was not rebased ontop of master. I did that now, and this is pushed to the repo. Thank you very much for your work!

PR #764, opened 2016-01-09
Add config option to disable backed error display

### LINKED ISSUE #763: Add config option to disable backend error displays entirely
I use python 3.4/3.5 for my projects (asyncio based).
Jedi raises exception on editing buffer with constructions like `yield from`, `async def`, `await` etc.

elpy reflects these errors by raising a buffer with bug report containing comprehensive report about the bug, but it's very annoying.

I use layout with two vertically separated buffers. As result the bug report appears in other (nonactive) buffer and focus is switched to it.

Is there an option to disable the annoying buffer flicker in case of Jedi errors?
I can live with partial autocomplete support.

Sorry, my knowledge of emacs list is close to zero and I have no time/intention to *contribute* into elpy and jedi. I just want to *use elpy with pleasure*.
So I just politely asking for elpy option which will make my life much easier.

ChillarAnand:
closes #763

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -310,6 +310,11 @@ edited instead. Setting this variable to nil disables this feature."
   :type 'string
   :group 'elpy)
 
+(defcustom elpy-disable-backend-error-display nil
+  "Non-nil if Elpy should disable backed error display."
+  :type 'boolean
+  :group 'elpy)
+
 ;;;;;;;;;;;;;
 ;;; Elpy Mode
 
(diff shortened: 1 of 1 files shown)


jorgenschaefer:
Awesome. You're quick! Thank you :-)

asvetlov:
@jorgenschaefer when do you expect publishing new release?

jorgenschaefer:
@asvetlov, at the end of this month (there should be a new Elpy release every three months). :-)

asvetlov:
sound good!
Thanks

PR #772, opened 2016-01-17
Add goto-definition function with grep fallback

el-ethan:
I found myself always calling `elpy-goto-definition`, and if it failed calling `elpy-rgrep-symbol` on the same symbol. To speed this process up a little bit, I wrote my own function that will try to run `elpy-goto-definition`, but will run `elpy-rgrep-symbol` on the symbol at point if the former function fails to find the definition. 

I am not sure if this belongs in Elpy, but it has been very useful for me, so I figured a pull request wouldn't hurt. I have bound it to `M-.`, but I didn't want to make that decision for the user, so I didn't add a default binding. Also, I am VERY new to elisp, so I would appreciate any feedback if something here is done in a problematic or non-idiomatic way. Thank you!

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -1647,6 +1647,16 @@ with a prefix argument)."
         (elpy-goto-location (car location) (cadr location))
       (error "No definition found"))))
 
+(defun elpy-goto-definition-or-rgrep ()
+  "Go to the definition of the symbol at point, if found. Otherwise, run `elpy-rgrep-symbol'.
+
+This first tries `elpy-goto-definition' on the symbol. If it fails to find the definition
+it passes the symbol, wrapped in a regexp for class/function definitions, off to `elpy-rgrep-symbol'"
+    (interactive)
+    (condition-case nil (elpy-goto-definition)
+        (error (elpy-rgrep-symbol
+                   (concat "\\(def\\|class\\)\s" (thing-at-point 'symbol) "(")))))
+
 (defun elpy-goto-location (filename offset)
   "Show FILENAME at OFFSET to the user."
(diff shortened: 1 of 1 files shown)


jorgenschaefer:
Hello, and thanks for the code!

I'm not sure if this is generally useful enough to have around, but it sure might be useful to some people! What do you think about adding this to the wiki instead, as user customization?

el-ethan:
That makes sense. Could you direct me to where it should go? I assume you are talking about this wiki: 
https://github.com/jorgenschaefer/elpy/wiki 
but I didn't see any section for user customization. Thank you!

jorgenschaefer:
There are some config snippets in the [FAQ](https://github.com/jorgenschaefer/elpy/wiki/FAQ), but you can also create a new file with a link from the main page. Maybe `Customizations` or `Contributions`?

el-ethan:
OK, I have my entry for the wiki ready. What is the workflow for contributing to the wiki? Should I just edit it directly through github, or should I fork, and make a pull request?

ChillarAnand:
You can  edit wiki directly.
May be create `Customizations` page and add your entry?

https://github.com/jorgenschaefer/elpy/wiki/_new

jorgenschaefer:
> What is the workflow for contributing to the wiki? Should I just edit it directly through github

Yes please, just edit it – if you create a new page, don't forget to link to it from the main page!

Thank you!

el-ethan:
OK, wiki page is done!

jorgenschaefer:
Thank you very much!  :-)

PR #773, opened 2016-01-17
Added elpy-goto-definition-other-window

### LINKED ISSUE #767: Add elpy-goto-definition-other-window
Hi, It would be really helpful to have analogue to `elpy-goto-definition` which would open definition in another window to read code side-by-side. For now one have to go to definition, then open another window, switch to it and then switch back to the initial buffer.

ChillarAnand:
closes #767

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -362,6 +362,7 @@ edited instead. Setting this variable to nil disables this feature."
     (define-key map (kbd "<M-right>") 'elpy-nav-indent-shift-right)
 
     (define-key map (kbd "M-.")     'elpy-goto-definition)
+    (define-key map (kbd "C-x 4 M-.")     'elpy-goto-definition-other-window)
     (define-key map (kbd "M-TAB")   'elpy-company-backend)
 
     map)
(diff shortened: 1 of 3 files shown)


jorgenschaefer:
Thanks for the contribution!

This makes it difficult to bind a key to the version that opens the location in another window. I think a separate command would be a better idea. As Emacs generally uses the `C-x 4` prefix, maybe `C-x 4 M-.` would be a good binding.

Also, can you write a test that actually tests this behavior? :-)

ChillarAnand:
```
(defun elpy-shell-switch-to-shell ()
  "Switch to inferior Python process buffer."
  (interactive)
  (pop-to-buffer (process-buffer (elpy-shell-get-or-create-process))))
```
Here `pop-to-buffer` is opening buffer in other window.

But when i try `(pop-to-buffer buffer)`, it is not opening buffer in some other window.
If I use `(pop-to-buffer buffer t)`, it is opening in other window but current buffer is getting switched to `buffer`.
Don't know what am i missing.

jorgenschaefer:
I'm not sure what the problem is, specifically. Could you say a bit more about exactly what you are doing, what you expect to happen, and what is happening instead?

ChillarAnand:
```
(defun foo (filename offset)
  (ring-insert find-tag-marker-ring (point-marker))
  (pop-to-buffer (find-file filename))
  (goto-char (1+ offset))
  (recenter 0))
```

When i evaluate `(foo bar baz)`

i expected



but this happened instead

jorgenschaefer:
`find-file` switches to the buffer of the file it opens – try using `find-file-noselect` instead?

ChillarAnand:
Ah, thank you, its  working  :-)

ChillarAnand:
Done!

jorgenschaefer:
Excellent, thank you! This looks good, could you squash the commits into one? I'll merge it then.

ChillarAnand:
Done!

jorgenschaefer:
Thank you very much for your contribution (all of them, actually, you are awesome!) :-)

maximvl:
Thanks!

Issue #771, opened 2016-01-17
Return correct form of Super() depending on Python version


gopar:
So in Python2 when you use `super`, it would be used like the following:
```python
super(ClassName, self).some_method()
```

But in Python3 you can simply do:
```python
super().some_method()
```
Well that's as far as I've seen. If I'm wrong feel free to correct me and close the issue.

But anyways, I modified the [func](https://github.com/jorgenschaefer/elpy/blob/fd1716883c36208078df30a578ff32b872751969/snippets/python-mode/.yas-setup.el#L49) responsible for helping the yas-snippet. Forgive my emacs lisp fu, but this is the general idea and it seems to work for me.
```lisp
(defun elpy-snippet-super-form ()
  "Return (Class, first-arg).method" 
  (let* ((defun-info (elpy-snippet-current-method-and-args))
         (class (nth 0 defun-info))
         (method (nth 1 defun-info))
         (args (nth 2 defun-info))
         (first-arg (nth 0 args))
         ;; Get the python version that is given when you run "python -v" in terminal. Either 2 or 3
         (py-version-num (substring (nth 1 (split-string (shell-command-to-string (concat elpy-rpc-python-command " -V")))) 0 1)))
    (if (string-match py-version-num "2")
        (format "(%s, %s).%s" class first-arg method)
      (format "().%s" method))))
```
If this seems like a good idea, maybe we could create a PR for it? Well let me know. Thanks for the work guys :)

jorgenschaefer:
Sounds good to me, thanks for the suggestion!

Would it be worthwhile to add an RPC call to Elpy to get the Python version for this?

gopar:
@jorgenschaefer not sure if I'm the right person to be asking :p 

Would you want a func just for getting the py version and then calling it into the other one? like this:

```lisp
(defun elpy-get-python-version ()
    "Return the python version"
  (nth 1 (split-string (shell-command-to-string (concat elpy-rpc-python-command " -V")))))

(defun elpy-snippet-super-form ()
  "Return (Class, first-arg).method" 
  (let* ((defun-info (elpy-snippet-current-method-and-args))
         (class (nth 0 defun-info))
         (method (nth 1 defun-info))
         (args (nth 2 defun-info))
         (first-arg (nth 0 args))
         (py-version-num (substring (elpy-get-python-version) 0 1)))
    (if (string-match py-version-num "2")
        (format "(%s, %s).%s" class first-arg method)
      (format "().%s" method))))
```

[...]

PR #778, opened 2016-01-20
Return correct form of "super" depending on Python version

gopar:
Previously when using the yasnippet "super", it will always return
"(class, arg).method" which was the Py2 way of doing it. In Py3, you
could simply do "().method" and Python will understand.

Solves #771


jorgenschaefer:
Thank you very much for your contribution! :-)

PR #823, opened 2016-03-07
Update documentation for moving by indentation

### LINKED ISSUE #753: Document usage of elpy-nav-forward-block etc. better
So I'm wondering why `elpy-nav-forward-block` does not fallback to forward-paragraph? 

It's very frustrating when I want to "scroll" through my file, but I get stuck.

I understand that it is was probably a choice, but perhaps it would make sense to fall back?

If people would like it, I could add the feature.

ChillarAnand:
closes #753


jorgenschaefer:
Thank you!

PR #825, opened 2016-03-11
elpy.el: fix compiler warning

stsquad:
When compiling elpy.el the byte compilers complains that insert-file is
intended for interactive use and insert-file-contents should be used
instead. As the function seems to have existed forever it seems pretty
safe to use.


jorgenschaefer:
Thank you very much! :-)

PR #836, opened 2016-03-17
Use "python -m pip" to install python modules instead of "pip".

xuhdev:
Currently "pip" is used to install missing modules in "elpy-config".
However, this may call the wrong version of pip, e.g. python3 often has
a pip executable "pip3" instead of "pip". This commit uses "python -m
pip" to call pip instead of the pip executable.

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -1130,9 +1130,10 @@ PyPI, or nil if that's VERSION."
                         ""))
 
          (command (cond
-                   ((executable-find "pip")
-                    (format "pip install %s%s%s"
-                            user-option upgrade-option python-package))
+                    ((= (call-process elpy-rpc-python-command
+                          nil nil nil "-m" "pip" "--help") 0)
+                      (format "%s -m pip install %s%s%s" elpy-rpc-python-command
+                        user-option upgrade-option python-package))
                    ((executable-find "easy_install")
                     (format "easy_install %s%s"
                             user-option python-package))
--- a/test/elpy-insert--pip-button-value-create-test.el
+++ b/test/elpy-insert--pip-button-value-create-test.el
@@ -12,8 +12,11 @@
     (mletf* ((executable-find
               (command)
               (cond
-                ((equal command "pip") t)
                 ((equal command "easy_install") t)))
+             (call-process
+              (program &optional infile destination display &rest args)
+               (when (and (equal program "python") (equal args '("-m" "pip"))))
+               0)
              (widget (widget-create 'elpy-insert--pip-button "test-module")))
       (should (string-match "pip" (widget-get widget :command)))
       (should (string-match "pip" (buffer-string))))))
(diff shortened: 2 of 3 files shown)


xuhdev:
It looks like the cause of the test failure is not this PR...

jorgenschaefer:
Yup. Looks like importmagic 1.7 is not compatible with Python 2.6 anymore – which is a reason as good as any to drop that compatibility as well. Feel free to remove (in a separate commit) the relevant line from `.travis.yml`.

Regarding your change, the ability to use `python -m pip` was only added around pip 6.1 I believe. At least Debian stable still ships with pip 1.5.6. Will this change really reduce problems more than it will cause them?

xuhdev:
I don't think this feature is new in pip 6.1---I'm using Debian Jessie right now (pip 1.5.6), and this feature works.

jorgenschaefer:
Ah. You are right, sorry. This is a problem of Python 2.6 which does not support `__main__.py`. And if we drop Python 2.6 support, that's not a problem. Your change sounds good to me.

elpy.el
```diff
@@ -1130,9 +1130,11 @@ PyPI, or nil if that's VERSION."
...
 
          (command (cond
-                   ((executable-find "pip")
-                    (format "pip install %s%s%s"
-                            user-option upgrade-option python-package))
+                    ((= (call-process elpy-rpc-python-command
+                          nil nil nil "-m" "pip" "--help") 0)
+                     (concat elpy-rpc-python-command
+                       (format " -m pip install %s%s%s"
+                               user-option upgrade-option python-package)))
```
jorgenschaefer: This should be `(format "%s -m pip install %s%s%s" elpy-rpc-python-command user-option ...)`

test/elpy-insert--pip-button-value-create-test.el
```diff
@@ -12,8 +12,10 @@
...
               (command)
               (cond
-                ((equal command "pip") t)
                 ((equal command "easy_install") t)))
+             (call-process
+              (program &optional infile destination display &rest args)
+               0)
              (widget (widget-create 'elpy-insert--pip-button "test-module")))
       (should (string-match "pip" (widget-get widget :command)))
       (should (string-match "pip" (buffer-string))))))
```
jorgenschaefer: This should probably check for `"python -m pip"`, no?

xuhdev: It looks unecessary to do so, ss there is only one call to call-process..

jorgenschaefer: I do not understand that argument. Why is it unnecessary to test whether the code actually calls the command you want it to call? The whole point of this patch is to change from "pip" to "python -m pip", no?

xuhdev:
@jorgenschaefer I've made the changes. Can you review it again? Thanks!

jorgenschaefer:
It seems we miscommunicated there. Don't worry, looks good to me, I'll clean it up. :-)

test/elpy-insert--pip-button-value-create-test.el
```diff
@@ -12,8 +12,11 @@
     (mletf* ((executable-find
               (command)
               (cond
-                ((equal command "pip") t)
                 ((equal command "easy_install") t)))
+             (call-process
+              (program &optional infile destination display &rest args)
+               (when (and (equal program "python") (equal args '("-m" "pip"))))
```
xuhdev: Oops, I forgot there is one more `--help` argument here... Could you add it, since it's already merged?

xuhdev: Seems that you have removed this line, never mind.

jorgenschaefer: Yes – that check is not doing anything, whether it works or not. The `should` statement should make sure that we use `python -m pip` now, not just `pip`. That's what I added now, too. :-)

Issue #690, opened 2015-10-13
Elpy should support dedicated Python processes


xin-jin:
It seems to me that currently all python buffers share a single python interpreter shell. Is it possible to associate different buffers to separate shells? In ess-mode, one can choose which shell to use (either an existing one or create a new one). I think it is very useful feature.

Thanks!

emacsway-qarea:
Have you tried rename current buffer before start new?

xin-jin:
@emacsway-qarea I did try. But that dissociate that shell with any buffer. What I want is run several different shells for different buffers at the same time, so that I can run two programs without their results messed up together.

jorgenschaefer:
Hello, and thanks for the question. :-)

You can do this using `C-u M-x run-python` – it will ask you which command to run, and then whether to make a "dedicated process". The latter is what you are looking for.

Does that solve the problem?

xin-jin:
@jorgenschaefer Thanks. Yes, I indeed can make a new dedicated process now. However, I find that the ``C-u C-u`` (send buffer to shell) behavior is then strange. It does send the buffer to the shell it associated with, but the "popped up" shell is the old one. 

That is, if I have buffer1 and buffer2, also shell1 and shell2, respectively. When I execute buffer2, the result is indeed printed on shell2, but shell1 rather than shell 2 pops up. Even when I manually split the frame to buffer2 and shell2, shell1 still pops up and messes up the frame:

![selection_002](https://cloud.githubusercontent.com/assets/6195346/10461860/467ebe34-7190-11e5-9b6d-33c4a3642736.png)

jorgenschaefer:
Oh, interesting. The buffer pop-up code is from Elpy, the rest is from python.el, and Elpy does not know about dedicated processes. I guess Elpy should support them at least rudimentarily.

Thanks for the report!

jonmoore:
I use the advice below for python-shell-get-process-name together with pyvenv-track-virtualenv so that I can have a mapping file->venv->process-name.  Based on limited tests this seems to work with elpy-shell-send-region-or-buffer, although it is obviously a bit fragile.
```elisp
(defun my-python-shell-get-process-name (orig-fun &rest args)
  "If `pyvenv-virtual-env-name' is set then return a buffer name
based on this and `python-shell-buffer-name', otherwise call
`python-shell-get-process-name'"
  (if pyvenv-virtual-env-name
[...]

PR #839, opened 2016-03-20
Fix dedicated shells support 

muahah:
I use a lot of dedicated shells and I had some troubles with that (c.f. issue #690).
I modified `elpy-shell-get-or-create-process` so that `elpy-shell-send-region-or-buffer` behave correctly.

I also added a new option to force the use of dedicated shells for each buffers
(to avoid having to launch `run-python` manually each time).
I ran some test and it also work with different virtual environments for each buffer.

I can run some more tests if the idea seems interesting enough.

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -153,6 +153,15 @@ uninteresting information, but if you rely on your modeline in other modes,
 you might want to keep it."
   :type 'boolean
   :group 'elpy)
+  
+(defcustom elpy-dedicated-shells nil
+  "Non-nil if Elpy should use dedicated shells.
+
+Elpy can use a unique python shell for all buffers and support
+manually started dedicated shells. Setting this option to non-nil
+force the creation of dedicated shells for each buffers."
+  :type 'boolean
+  :group 'elpy)
 
 (defcustom elpy-rpc-backend nil
   "Your preferred backend.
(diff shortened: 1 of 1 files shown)


jorgenschaefer:
Very nice, thank you for the contribution! This is definitely interesting :-)

It looks like you could likely simplify the code a lot by passing `elpy-dedicated-shells` directly to `python-shell-get-process-name` and only have to call that once – the remaining code looks very much alike. Or am I missing something?

muahah:
If you don't want to use dedicated shells by default (`elpy-dedicated-shells`to `nil`), but you need a dedicated shell for some particular scripts (launched manually with `C-u run-python`), `elpy-shell-get-or-create-process` need to check if the buffer have a dedicated shell (check `dedproc`) and if not, use the default shell (`proc`). 
So in this particular case, you need `proc` and `dedproc` at the same time.

But maybe it could be better to get `proc` and `dedproc` after testing the `elpy-dedicated-shells` option, so that `proc` is not gotten for nothing when `elpy-dedicated-shells` is `t` ?

jorgenschaefer:
I have a hard time knowing how that would look like – use whichever approach leads to the most simple and most straightforward readable code. :-)

muahah:
I will try to find something not too ugly.
:)

muahah:
I couldn't find a nicer way of doing it, and it seems to work.
The dedicated shells buffer names can be annoyingly long due to the absolute path, but I didn't find any nice (and simple) solution...

jorgenschaefer:
Looks good to me! Can you squash the commits in this PR into one? I'll merge it then :-)

muahah:
Done !

jorgenschaefer:
Excellent, thank you very much for your contribution!

muahah:
I use elpy on a daily basis, it's the least I can do.
Excellent package by the way, It really gave me hope to use python efficiently within emacs.

PR #865, opened 2016-04-12
Enable highlight-indentation-current-column-mode in the highlight-indentation module

xuhdev:


jorgenschaefer:
Hello and thank you for the contribution!

I can totally see how this mode can come in handy. Seeing related code, especially in slightly longer functions, is very helpful. Sadly, I found the highlighting of the current column to be confusing more often than helpful – it seems to jump around a lot while moving, which distracts me from actually keeping focused on the piece of code I'm reading. And long pieces of code indented deeply where this would show its best benefit actually should be refactored.

So, considering how distracting I found it, and how easy it is to enable for those who like it (or when they want it), I think I'd rather not enable it by default. Sorry.

PR #892, opened 2016-05-26
Set python-check-command correctly

### LINKED ISSUE #790: Set python-check-command correctly even if flake8 is not installed on Emacs startup
`elpy-module-flymake` only sets `python-check-command` to `"flake8"` when flake8 is actually in the path during Emacs startup, which is not the case if flake8 is installed only in virtualenvs and not available on Emacs start.

Elpy should be smarter than that.

See #789 for a discussion.

ChillarAnand:
closes #790

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -340,6 +340,11 @@ edited instead. Setting this variable to nil disables this feature."
   :type 'boolean
   :group 'elpy)
 
+(defcustom elpy-syntax-check-command "flake8"
+  "The command to use for `elpy-check'."
+  :type 'string
+  :group 'elpy)
+
 ;;;;;;;;;;;;;
 ;;; Elpy Mode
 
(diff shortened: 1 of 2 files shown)


jorgenschaefer:
Hm. No, I don't think using the flymake module like that is the right approach. Users might have disabled the module.

I think the right thing to do here is to have a separate Elpy setting for the check command that users can override. Make it default to "flake8", and set `python-check-command` to that during `elpy-enable`.

ChillarAnand:
Ah, I see. Updated it.

jorgenschaefer:
Thank you!

Do you think using that as the package name is a good idea? Someone could set it to e.g. `/foo/bar/baz/bin/flake8` instead, and that would break the `elpy-config` display.

ChillarAnand:
I was thinking, if user sets `pylint` and activates a venv where it is not available, that pip button might help.

I haven't thought of setting pylint as `/foo/pylint`. Added a test case for it.

jorgenschaefer:
Looks good to me. Let me know if you think this can be merged. :-)

ChillarAnand:
Looks good to me too :-)

jorgenschaefer:
Thank you!

PR #905, opened 2016-06-04
Improve cursor positioning while format

### LINKED ISSUE #824: elpy-rpc--default-error-callback: Elpy error: yapf not installed
I'm trying use `C-c C-r f` to run yapf on a region, but when I try this, Emacs responds with `Elpy error: yapf not installed`. (The subject lists the message as it appears in the `*Messages*` buffer.)

The relevant file uses a virtualenv, `elpy-config` reports the following:

```
Virtualenv........: python3 (/home/joost/src/mal/python3)
RPC Python........: 3.4.3 (/home/joost/src/mal/python3/bin/python)
Interactive Python: ipython (/home/joost/src/mal/python3/bin/ipython)
Emacs.............: 25.0.92.1
Elpy..............: 1.11.0
Jedi..............: 0.9.0
Rope..............: Not found (0.10.3 available)
Importmagic.......: 0.1.5
Autopep8..........: 1.2.2
Yapf..............: 0.6.3
Syntax checker....: flake8 (/home/joost/src/mal/python3/bin/flake8)
```

Which looks to me like everything should be fine. Note that `yapf` is only installed in the virtualenv, though, not system-wide. Should I install `yapf` system-wide as well? Won't that produce the wrong results, though, given that the system Python is version 2.7, whereas I use 3.4 in the virtualenv.

ChillarAnand:
closes #824

### DIFF
--- a/test/elpy-yapf-fix-code-test.el
+++ b/test/elpy-yapf-fix-code-test.el
@@ -15,3 +15,27 @@
                       (buffer-be
                        "x = 1"
                        "y = 2_|_"))))))
+
+(ert-deftest elpy-yapf-fix-code-in-region-should-retain-line-and-column ()
+  (let* ((pyversion (getenv "TRAVIS_PYTHON_VERSION"))
+         (yapf-not-supported (or (string< pyversion "2.7")
+                                 (and (not (string< pyversion "3.0"))
+                                      (string< pyversion "3.4")))))
+    (unless yapf-not-supported
+      (elpy-testcase ()
+                     (set-buffer-string-with-point
+                      "_|_y =  2"
+                      "z = 3"
+                      "x = 3"
+                      )
+                     (elpy/mark-region (point) 13)
+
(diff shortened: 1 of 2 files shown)


test/elpy-yapf-fix-code-test.el
```diff
@@ -15,3 +15,28 @@
...
+
+                     (elpy-yapf-fix-code)
+
+                     (should (not mark-active))
+                     (should
+                      (buffer-be
+                       "x  =  1"
+                       "y = 2"
+                       "z = 3"
+                       "_|_"
```
ChillarAnand: `      (buffer-contents-differ "x  =  1\ny = 2\nz = 3\n_|_\n" "x  =  1\ny = 2\nz = 3\n_|_")))
`
 Any idea why `\n` is getting added in the end again? @jorgenschaefer

jorgenschaefer:
Thank you very much!

PR #951, opened 2016-08-14
Amend "no backend available" message ...

offby1:
... with a pointer to the "Should I use Rope or Jedi?" FAQ


jorgenschaefer:
Thank you!

PR #958, opened 2016-08-24
Add basic Django support

gopar:
You can now parse through `manage.py` (If it is found, otherwise it
falls back to `django-admin.py`) for commands to run and emacs will
provide completion assistance.

Another function is to automatically start the development server
command, `runserver`, with default variables that are set to
`127.0.0.1`(for ip address) and `8000` (for port). Each Django project
has the ability of starting their own dev server.

One question I have regarding this is if we should also allow changing
the command for starting the development server, `runserver`? I know
other people like `runserver_plus` from the `django-extensions` module.

Commit also introduces a new dependency, `s`, for string
manipulation. Currently, it is only used on one line for trimming
whitespace. Should the commit introduce this new dep or simply create
a function that does this?

Forgot to mentioned: User can specify extra arguments by starting the command with a prefix `C-u` or if they'd like to always prompt for extra arguments, there is a variable for that as well.

### DIFF
--- a/Cask
+++ b/Cask
@@ -10,6 +10,7 @@
 (depends-on "highlight-indentation" "0.5.0")
 (depends-on "pyvenv" "1.2")
 (depends-on "yasnippet" "0.8.0")
+(depends-on "s" "1.10.0")
 
 (development
  (depends-on "dash")
--- a/docs/ide.rst
+++ b/docs/ide.rst
@@ -359,3 +359,24 @@ Elpy supports various forms of refactoring Python code.
    options depending on the context.
 
 .. _rope: https://github.com/python-rope/rope
+
+
+Django
+======
+
+Elpy has basic Django support such as parsing either `manage.py` or `django-admin.py` (If it
+does not find `manage.py` it falls back to `django-admin.py`) for command completion assistance.
+Can also start `runserver` automatically and you can give an ip address and port.
+
+.. command:: elpy-django-command
+   :kbd: C-c C-x c
+
+   Choose what command you'd like to run via `django-admin.py` or `manage.py`.
+
+.. command:: elpy-django-runserver
+   :kbd: C-c C-x r
+
+
+   Start the development server command, `runserver`. Default arguments are `127.0.0.1` for
+   ip address and `8000` for port. These can be changed via ``elpy-django-server-ipaddr`` and
+   ``elpy-django-server-port``.
--- a/elpy-django.el
+++ b/elpy-django.el
@@ -40,14 +40,14 @@ Best to set it to full path to 'manage.py' if it's available."
 (defcustom elpy-django-server-ipaddr "127.0.0.1"
   "What address Django will use when running the dev server."
   :type 'string
-  :type 'stringp
+  :safe 'stringp
   :group 'elpy)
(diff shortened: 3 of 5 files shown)


jorgenschaefer:
Wow, thank you! Some quick notes:

- `C-c C-x` is a pretty big prefix key. I'm not sure I like it, but I have no good better idea there. `x` for general `eXecute` might be ok.
- I have no problem with depending on `s.el`. Newer Emacsen come with `string-trim`, but we'd still need a backwards-compatible alias.
- You can just provide a runserver command variable that is `"runserver"` by default
- Thank you for making this a separate file! And for adding tests! You are awesome. :-)

I'll look closer at this when I have a dedicated Elpy month again.

gopar:
Hmm the only other keybinding that comes to mind is `C-c C-w` (web) but if you want to add support for different web frameworks (Flask, Pyramid, etc) it might get a little weird (Well at least that's my opinion).

Let me know what keybinding you'd like it to be.

I'll add the extra runserver variable.

vindarel:
Hi,
Looks cool. Any reason why you wanted to add that to elpy and not to [django-mode](https://github.com/myfreeweb/django-mode) ? 
To this mode I added the completion of any manage.py commands a few weeks ago. I chose to always prompt for the final command. It's not a big deal, if we're ok we just press Enter right away. That would prevent the burden of having a list of django commands that need a prompt (`elpy-django-commands-with-req-arg` and else), which can be uncomplete because of commands coming from extensions or just custom commands.

gopar:
@vindarel hey
I added the PR to elpy since django-mode stated that it's no longer maintained and when I did the first initial commit of this PR, your changes were not yet there. Also, the last few commits have been PR's sprinkled over the last couple of years. Didn't seem like a vivid project at the time.

As for the always asking for a prompt, I kinda like that idea since we don't know if there will be external commands. I tried getting around that by making the list unique for each project, and the user can modify it as they need it. I'm not sure if always asking for prompt would be good though, since as a user I'd kinda hate always having to press an extra Enter. I'd rather leave it up to @jorgenschaefer since it's his project \o/

jorgenschaefer:
Indeed, extra prompts are annoying when they can be avoided. Thank you *very* much for this contribution!

gopar:
Hey @jorgenschaefer were you able to test out this branch? I kinda left it unchecked for a while since I wasn't doing any Django for a bit. I wouldn't want to be the guy to introduce a broken feature :(

jorgenschaefer:
Nope. No worries, if it broke something, it's my fault. :-D

gopar:
Alright \o/

Also, were you fine with the keybinding of `C-c C-x`? Better to change it now, than wait later on \o/

jorgenschaefer:
I could not think of a better key binding. If we come up with something else that needs that prefix later, we can still extend it then. :-)

vindarel:
Hi ! I find `C-c C-x` dangerous, because we can easily mistape it and quit emacs (`C-x C-c`).

gopar:
@vindarel Hmm any suggestions? The only other keybinding that I was thinking about was `C-c C-j`

Hmm personally, I've never mistyped the keybinding but that could just be me

PR #971, opened 2016-09-08
Send project root to importmagic

ChillarAnand:
importmagic can read configuration from setup.cfg which will be
available in project root. So, elpy should send project root to it.

https://github.com/alecthomas/importmagic/pull/39

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -2240,9 +2240,11 @@ prefix argument is given, prompt for a symbol from the user."
   "Prompt to import thing at point, show possbile imports and add selected import."
   (interactive (elpy-importmagic--add-import-read-args))
   (unless (equal statement "")
+    (elpy-project-root)
     (let* ((res (elpy-rpc "add_import" (list buffer-file-name
                                              (elpy-rpc--buffer-contents)
-                                             statement))))
+                                             statement
+                                             elpy-project-root))))
       (elpy-buffer--replace-block res))))
 
 (defun elpy-importmagic-fixup ()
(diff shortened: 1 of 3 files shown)


elpy/impmagic.py
```diff
@@ -68,8 +68,9 @@ def sort_key(item):
         return ["from %s import %s" % (mod, var) if var else "import %s" % mod
                 for (_, mod, var) in scores]
 
-    def add_import(self, source, statement):
-        imports = importmagic.importer.Imports(self.symbol_index, source)
+    def add_import(self, source, statement, project_root):
```
gopar: Shouldn't `project_root=None` be in the function definition instead of just `project_root`? That way in line 73 you can just pass `project_root` without setting it to `None`.

elpy.el
```diff
@@ -2261,7 +2261,8 @@ Also sort the imports in the import statement blocks."
           (elpy-importmagic-add-import (car choice))))))
   ;; now get a new import statement block (this also sorts)
   (let* ((res (elpy-rpc "remove_unreferenced_imports" (list buffer-file-name
-                                                            (elpy-rpc--buffer-contents)))))
+                                                            (elpy-rpc--buffer-contents)
+                                                            elpy-project-root))))
```
jorgenschaefer: Shouldn't this use (elpy-project-root) to calculate the root if there is none yet?

galaunay:
We should wait to see the outcome of #1230 before working on this

gopar:
Closing, since support for impormagic has been dropped

PR #976, opened 2016-09-12
One-line python requirements install line added

Thalmann:
When i was installing elpy i was missing this - i find it easier to copy the line than typing it all out or getting the repo and installing via the requirements file :)
Up to you ofc.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -36,7 +36,7 @@ First, install the required Python packages:::
   # and yapf for code formatting
   pip install yapf
 
-One-line install: **sudo pip install rope jedi flake8 importmagic autopep8 yapf**
+One-line install: **pip install rope jedi flake8 importmagic autopep8 yapf**
 
 Evaluate this in your ``*scratch*`` buffer:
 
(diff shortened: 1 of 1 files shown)


gopar:
Shouldn't `pip3` just be `pip`? Since people might want to use Elpy for py2 dev and not py3. Also, if you're in a virtual env then doing pip3 might be install packages to the global system python instead of your virtual one.

`sudo` might also be a little extra since not everyone is in a linux machine (Windows) and can confuse people (beginners), plus some people might not have sudo powers and just install in user directory.

Best to keep it generic but this is just my point of view. Up to @jorgenschaefer

Thalmann:
I totally agree with you, i thought it was only python 3 for some reason :)

jorgenschaefer:
Hello, and thanks for the contribution! I like this! Could you remove `rope` and `yapf` (they do the same thing as `jedi` and `autopep8`) and squash the commits into one? I'll be happy to merge then :-)

Thalmann:
Fixed here: https://github.com/jorgenschaefer/elpy/pull/982

PR #989, opened 2016-09-21
Source yapf style file from elpy-project-root or default-directory.

jonmoore:
The suggestion is to do this rather than using os.getcwd(), which does not pick up project-specific  yapf style settings, similarly to #971 for importmagic.

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -2283,19 +2283,24 @@ Also sort the imports in the import statement blocks."
 (defun elpy-yapf-fix-code ()
   "Automatically formats Python code with yapf."
   (interactive)
-  (elpy--fix-code-with-formatter "fix_code_with_yapf"))
+  (elpy--fix-code-with-formatter "fix_code_with_yapf"
+                                 (or (expand-file-name (elpy-project-root))
+                                     default-directory)))
 
 (defun elpy-autopep8-fix-code ()
   "Automatically formats Python code to conform to the PEP 8 style guide."
   (interactive)
   (elpy--fix-code-with-formatter "fix_code"))
 
-(defun elpy--fix-code-with-formatter (method)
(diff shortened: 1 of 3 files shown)


jorgenschaefer:
Hello, and thank you for the contribution! This looks very good. Do you think there could be tests for this functionality, especially for the Python side?

In any case, could you squash the commits? I'll be happy to merge. :-)

landakram:
@jorgenschaefer GitHub now lets folks squash commits on PR merge. Given that, can you merge it and squash when you merge? Otherwise, let's hope @jonmoore sees this :)

Signed, someone who benefits from this PR

galaunay:
I can add tests for this, 
but after taking a look at the yapf documentation, it says:
>   --style STYLE         
specify formatting style: either a style name (for
example "pep8" or "google"), or the name of a file
with style settings. The default is pep8 unless a
.style.yapf or setup.cfg file located **in one of the
parent directories of the source file** (or current
 directory for stdin)

So is this really necessary, as yapf should already pick up the style file present in the project root directory ?

d8660091:
@galaunay Nope. I ran into this problem the other day. It turns out that elpy tries to pick up styles at my root directory '/' in my machine. I know it's caused by os.getcwd() but I didn't figured out how cwd is set to '/'.

galaunay:
@d8660091 I see it now.
yapf is called from the rpc, for which the current directory is `/`.

I just tested and took a look at the PR, it seems fine.
However, It may worth indicating somewhere that Elpy is now looking for yapf style file in the project root.

I will try to add this and merge the PR in the following days.

jorgenschaefer:
You people are all awesome. That is all. :-D

galaunay:
I made another PR for that.
Closing this one !

PR #994, opened 2016-09-23
Update README.rst

RafaelCatrou:


jorgenschaefer:
Hello, and thank you for the contribution! There is no need to install both Rope _and_ Jedi, and I'm fine with suggesting Jedi over Rope in that one-line installation. The same applies to autopep8 and yapf. Hence the line is fine as is, I would say.

(In general, try to phrase your commit messages a bit more clearly as to what they do, and preferably also why they do it. http://chris.beams.io/posts/git-commit/ has a good intro. :-))

Issue #931, opened 2016-07-09
Running unit test with bad project root (Django)


vindarel:
Hi !

So today I finally discovered this elpy feature.

First I had to dive in the code to see and edit the following variable `(setq elpy-test-django-runner-command '("./manage.py" "test"))`.

Then I tried to run my unit test with `M-x elpy-test-django-runner` but it failed because the project root was one directory too up.

I edited elpy.el's `elpy-test-at-point` like so

```
    ;; (let* ((top (elpy-library-root))
  (let* ((top (elpy-project-find-projectile-root))
```

and it worked :)

So I don't know if it's just me, but to me this elpy command is broken. Is that so, or can I fix the project root by another mean ?

Using elpy 20160613.

thanks !

jorgenschaefer:
Hello, and thanks for the report! Your problem here is that you are using `manage.py` to run the tests, and then even use a relative path to run it. That means that the way Elpy finds project roots breaks. Your first "fix" there, changing `elpy-test-django-runner-command`, is likely the wrong way to solve the actual problem you have.

Elpy uses `django-admin.py` instead of `manage.py` for this very reason: The location of `manage.py` is not uniform and easy to find. Instead of doing those changes you did there, you should just set the `DJANGO_SETTINGS_MODULE` environment variable to your settings module, and everything should work out of the box.

Does this help?

vindarel:
mmh, it helps to know that we must change this environment variable, but all isn't that clear to me.

> That means that the way Elpy finds project roots breaks

what's the relation ? Projectile finds the project root and Elpy runs a correct command, relative pathname or not.

I understand now the use of django-admin.py with the environment variable, but it just looks more complicated to me. We must set the variable and change it for every project right ? Not with the way I did.

Actually I couldn't run my test with your method even on the terminal, I have an "Import by filename is not supported" error…

Regards

jorgenschaefer:
> Projectile finds the project root and Elpy runs a correct command, relative pathname or not.

You assume that all projects have a `manage.py` at the project root. That's actually not all that common. Great that it works for you.

You also assume that test runners should be run from the project root, and not the library root. That's also not always the right thing.

I am not saying that "your way" is wrong or anything, I am just pointing out that just because it works for you does not mean that it works for everyone. :-)

> Actually I couldn't run my test with your method even on the terminal, I have an "Import by filename is not supported" error…

Could you provide the full command being run and the full error? Thank you!

vindarel:
[...]

PR #999, opened 2016-09-30
Option to run Django tests with manage.py

vindarel:
_"…setting the project root accordingly,
in order not to set the environnement variable DJANGO_SETTINGS_MODULE
for every project (but instead change of virtual env)."_

This PR does not change elpy's default behavior.

It adds a variable to tell elpy to run django tests with `manage.py` instead of `django-admin.py`. 
The thing is that we can't just set `elpy-test-django-runner-command` to our manage.py command because `elpy-test-at-point` needs to return the project root, not the library root.

Benefits: it allows a user to run tests quicker, with no venv or environment variable configuration. We just have to change of virtual env, which is easily done with other packages ([virtualenvwrapper.el](https://github.com/porterjamesj/virtualenvwrapper.el) and others). 

Because from my experience, I wasn't able to run django tests when I tried elpy. Now, I can and I hope it's easier for others too.

See https://github.com/jorgenschaefer/elpy/issues/931

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -315,6 +315,18 @@ edited instead. Setting this variable to nil disables this feature."
   :type '(repeat string)
   :group 'elpy)
 
+(defcustom elpy-test-django-runner-manage-command '("manage.py" "test"
+                                                    "--noinput")
+  "The command to use for `elpy-test-django-runner' in case we want to use manage.py."
+  :type '(repeat string)
+  :group 'elpy)
+
+(defcustom elpy-test-django-with-manage nil
+  "Set to nil, elpy will use `elpy-test-django-runner-command',
+set to t elpy will use `elpy-test-django-runner-manage-command' and set the project root accordingly."
+  :type 'boolean
+  :group 'elpy)
+
 (defcustom elpy-test-nose-runner-command '("nosetests")
   "The command to use for `elpy-test-nose-runner'."
   :type '(repeat string)
@@ -2050,17 +2062,28 @@ This requires Python 2.7 or later."
 (put 'elpy-test-discover-runner 'elpy-test-runner-p t)
 
 (defun elpy-test-django-runner (top file module test)
-  "Test the project using the Django discover runner.
+  "Test the project using the Django discover runner,
+or with manage.py if elpy-test-django-with-manage is true.
 
 This requires Django 1.6 or the django-discover-runner package."
   (interactive (elpy-test-at-point))
   (if module
       (apply #'elpy-test-run
              top
-             (append elpy-test-django-runner-command
-                     (list (if test
-                               (format "%s.%s" module test)
-                             module))))
+             (append
+              ;; if we want to use manage.py, get the root directory where it is.
+              (if elpy-test-django-with-manage
+                  (append (list (concat (expand-file-name
+                                         (locate-dominating-file
+                                          (elpy-project-root)
+                                          (car elpy-test-django-runner-manage-command)))
+                                        (car elpy-test-django-runner-manage-command)))
+                          (cdr elpy-test-django-runner-manage-command))
+                ;; or the default:
+                elpy-test-django-runner-command)
+              (list (if test
+                        (format "%s.%s" module test)
+                      module))))
     (apply #'elpy-test-run
            top
            elpy-test-django-runner-command)))
--- a/docs/ide.rst
+++ b/docs/ide.rst
@@ -291,8 +291,15 @@ advantages this brings.
 
    This changes the current test runner. Elpy supports the standard
    unittest discovery runner, the Django discovery runner, nose and
-   py.test. You can also write your own, as described in
-   :ref:`Writing Test Runners`.
+   py.test. You can also write your own, as described in :ref:`Writing
+   Test Runners`.
+
+   Note on Django runners: by default, elpy runs Django tests with
+   :kbd:`django-admin.py`. You must set the environnement variable
+   :envvar:`DJANGO_SETTINGS_MODULE` accordingly. Alternatively, you can set
+   **elpy-test-django-with-manage** to **t** in order to use your
+   project's :kbd:`manage.py`. You then don't need to set the environnement
+   variable, but change virtual envs (see `virtualenvwrapper.el`_).
 
 This enables a good workflow. You write a test and use :kbd:`C-c C-t`
 to watch it fail. You then go to your implementation file, for example
@@ -305,6 +312,8 @@ For an even more automated way, you can use `tdd.el`_, which will run
 your last compile command whenever you save a file.
 
 .. _tdd.el: https://github.com/jorgenschaefer/emacs-tdd/
+
+.. _virtualenvwrapper.el: https://github.com/porterjamesj/virtualenvwrapper.el
 
 
 Refactoring


jorgenschaefer:
Hello, and thanks for the contribution – especially improving the documentation at the same time is excellent!

The changes look basically ok to me. A few comments, though:

You are using `elpy-project-find-projectile-root` directly, which creates a hard projectile dependency. There is no need for this as far as I can see, you can just use `elpy-project-root`.

Or even better yet, you could use `locate-dominating-file` and search for a `manage.py` file directly.

Do you see a problem with that?

vindarel:
Rich idea, I added locate-dominating-file (nice method !) with elpy-project-root, expanding the file name just to be sure.

Indeed for the doc :)

(I rebased the PR)

jorgenschaefer:
Heh, that was fast :-) I'm afraid this won't work – the first argument to `locate-dominating-file` is the file name, the second is the directory. It also returns a file name, not a directory name.

You should use `locate-dominating-file` in elpy-test-django-runner to find the executable to run, not to change any top directory or anything.

vindarel:
I pushed the change with elpy-project-root.

I'm afraid I don't understand the other suggestion then.
Maybe can you comment on the specific line ?
(to me it would be `locate-dominating-file start-directory file-name-to-look-for`, which returns `~/project/`)

elpy.el
```diff
@@ -1989,7 +2001,9 @@ directory is not nil."
       (progn
         (save-some-buffers)
         (list (elpy-library-root) nil nil nil))
-    (let* ((top (elpy-library-root))
+    (let* ((top (if elpy-test-django-with-manage
+                  (elpy-project-root)
+                 (elpy-library-root)))
```
jorgenschaefer: You should not have to change this function at all.

elpy.el
```diff
@@ -2050,14 +2064,17 @@ This requires Python 2.7 or later."
...
 This requires Django 1.6 or the django-discover-runner package."
   (interactive (elpy-test-at-point))
   (if module
       (apply #'elpy-test-run
              top
-             (append elpy-test-django-runner-command
+             (append (if elpy-test-django-with-manage
+                         elpy-test-django-runner-manage-command
+                       elpy-test-django-runner-command)
                      (list (if test
```
jorgenschaefer: Instead, use `locate-dominating-file` to get the full path to `manage.py` to use instead of `elpy-test-django-runner-command`.

vindarel: So ! I don't clearly see what you suggest. I tried:
```lisp 
(list (concat 
    (expand-file-name (locate-dominating-file (elpy-project-root) "manage.py"))
    ;; returns a directory path, so I must append the file name again.
    ;; This snippet will get more verbose for the need of using the variable `elpy-test-django-runner-manage-command`. 
     "manage.py"))
```
the final command is:

    /home/.../parentdir/projectname/manage.py projectname.myapp.tests-dir.etc

the projectname in the last part is wrong. I suspect it's there because the `top` variable is at `parentdir`.

That's where I am. Anyway, I feel this is a lot of machinery, not necessarily cleaner than keeping the first "if" above, in `elpy-test-at-point` :]

jorgenschaefer: `(concat (locate-dominating-file (buffer-file-name) "manage.py") "manage.py")` will give you the full path to a `manage.py` which you can use as the command to run. This is a lot cleaner than your solution, as it will work in a lot more situations. Please also think about what "the command to run" means, as that will simplify the rest of your code.

vindarel:
Ahah ! yes ?

elpy.el
```diff
@@ -2050,14 +2063,19 @@ This requires Python 2.7 or later."
...
       (apply #'elpy-test-run
-             top
-             (append elpy-test-django-runner-command
+             (if elpy-test-django-with-manage
+                 (expand-file-name (locate-dominating-file (elpy-project-root) "manage.py"))
+               top)
+             (append (if elpy-test-django-with-manage
+                         elpy-test-django-runner-manage-command
+                       elpy-test-django-runner-command)
                      (list (if test
```
jorgenschaefer: - The _directory_ in which to run `elpy-test-run` should *probably* be the directory of `manage.py`, no?
- The _command_ to run should be the full path to manage.py

vindarel:
Hi ! So, I know I didn't rebase the last two commits, it's on purpose for your review. Now it looks to me we are closer to what you ask: the code is changed in only one place. Hope it's good. It's working though :]

elpy.el
```diff
@@ -315,6 +315,18 @@ edited instead. Setting this variable to nil disables this feature."
   :type '(repeat string)
   :group 'elpy)
 
+(defcustom elpy-test-django-runner-manage-command '("manage.py" "test"
+                                             "--noinput")
```
jorgenschaefer: Why is this on a new line? And if it is, why is it not indented as per normal Emacs Lisp rules (just use `TAB`)?

elpy.el
```diff
@@ -315,6 +315,18 @@ edited instead. Setting this variable to nil disables this feature."
...
+  "The command to use for `elpy-test-django-runner' in case we want to use manage.py."
+  :type '(repeat string)
+  :group 'elpy)
+
+(defcustom elpy-test-django-with-manage nil
+  "Set to nil, elpy will use `elpy-test-django-runner-command',
+   set to t elpy will use `elpy-test-django-runner-manage-command' and set the project root accordingly."
```
jorgenschaefer: No spaces at the beginning of the line here.

elpy.el
```diff
@@ -1989,7 +2001,8 @@ directory is not nil."
       (progn
         (save-some-buffers)
         (list (elpy-library-root) nil nil nil))
-    (let* ((top (elpy-library-root))
+    (let* ((top
+            (elpy-library-root))
```
jorgenschaefer: No need to add a line break here.

elpy.el
```diff
@@ -2050,17 +2063,27 @@ This requires Python 2.7 or later."
 (put 'elpy-test-discover-runner 'elpy-test-runner-p t)
 
 (defun elpy-test-django-runner (top file module test)
-  "Test the project using the Django discover runner.
+  "Test the project using the Django discover runner,
+   or with manage.py if elpy-test-django-with-manage is true.
```
jorgenschaefer: No need for spaces at the beginning of the line here.

elpy.el
```diff
@@ -2050,17 +2063,27 @@ This requires Python 2.7 or later."
...
-             (append elpy-test-django-runner-command
-                     (list (if test
-                               (format "%s.%s" module test)
-                             module))))
+             (append
+              ;; if we want to use manage.py, get the root directory where it is.
+              (if elpy-test-django-with-manage
+                         (append
+                          (list (concat
+                           (expand-file-name (locate-dominating-file (elpy-project-root) (car elpy-test-django-runner-manage-command)))
```
jorgenschaefer: On the other hand, this line could really be split up on multiple lines to make it clearer what is happening here. (General rule of thumb, a line break can be useful after a number of consecutive closing parens).

elpy.el
vindarel: sure, because of a search&replace.

jorgenschaefer:
Thank you for your contribution!

vindarel:
Awesome, thanks to you too :)

PR #1000, opened 2016-09-30
Fix #984

### LINKED ISSUE #984: Make C-c C-z remember the location jumped from to jump back to
Hi,

When I'm in the inferior python buffer and press C-c C-z, it should jump to the last viewed python script buffer, but it seems to jump to any buffer I last viewed. The elisp code also proves that:
```
(defun elpy-shell-switch-to-buffer ()
  "Switch from inferior Python process buffer to recent Python buffer."
  (interactive)
  (pop-to-buffer (other-buffer (current-buffer) 1)))
```
I think it would be easy to filter the buffer list and only pop to the first item in the list which is python mode, or something like that. However, I'm not familiar with elisp, so would you mind providing a refined version of this function?

Many thanks.

gopar:
When calling the py shell from within a python file, it will
save the python file location so when you call `C-c C-z` from the shell,
it will properly return you to the file that initiated that shell call.

This goes the route of option(3) as discussed in #984

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -347,6 +347,8 @@ edited instead. Setting this variable to nil disables this feature."
 
 ;;;;;;;;;;;;;
 ;;; Elpy Mode
+(defvar elpy--shell-last-py-buffer nil
+  "Help keep track of python buffer when changing to pyshell.")
 
 (defvar elpy-refactor-map
   (let ((map (make-sparse-keymap "Refactor")))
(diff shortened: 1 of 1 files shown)


jorgenschaefer:
Excellent, thank you!

jorgenschaefer:
Also, congratulations on the #1000 there! :-D

PR #1032, opened 2016-11-27
Add functions to kill shell(s)

muahah:
## Motivation
With the use of dedicated shells, I end up having a lot of active python shells (potentially using a lot of memory...). 
As it can be time-consuming to search and kill each python shell, I wrote a function to do it automatically.

## New functions
`elpy-shell-kill` : kill the current python shell (and the python shell buffer, if asked).
`elpy-shell-kill-all` : kill all the dedicated python shells (and their buffers, if asked).

### DIFF
--- a/docs/ide.rst
+++ b/docs/ide.rst
@@ -183,6 +183,18 @@ easier to send code snippets over.
    Similar to :kbd:`C-c C-c`, this will send the code of the current
    top level class or function to the interactive Python process.
 
+.. command:: elpy-shell-kill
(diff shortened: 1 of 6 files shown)


jorgenschaefer:
Very nice! Also, nice use of `loop` :-D

Could you squash the commits into one, please? I'll be happy to merge.

muahah:
Done,
thanks again for elpy, definitely on my top ten emacs package !

jorgenschaefer:

PR #1033, opened 2016-11-27
Add aliases to elpy importmagic

muahah:
## Motivation
I use a lot a aliases while coding in python.
As importmagic handles aliases, it seems a good idea for the  `elpy-importmagic-*` functions to handle them as well.

## Improvements
- `elpy-importmagic-add-import` with non-nil `ask-for-alias` optional argument now ask for the wanted alias after choosing the import statement.
- `elpy-importmagic-fixup` now gathers unresolved module names (supposed to be aliases), ask for their real names and add import statements in consequence.

## Working (simple) example
```python
HOME = r"/home/ben"

x = np.linspace(0, 1, 1000)
y = x**2
plt.figure()
plt.plot(x, y, marker=markers.TICKDOWN)
plt.title(time.time)
plt.show(block=False)
plt.savefig(join(HOME, "test-elpy/figure.svg"))
```
give, after `elpy-importmagic-fixup` and some prompting: 
```python
import time
from os.path import join

import numpy as np
from matplotlib import markers, pyplot as plt


HOME = r"/home/ben"

x = np.linspace(0, 1, 100)
y = x**2
plt.figure()
plt.plot(x, y, marker=markers.TICKDOWN)
plt.title(time.time)
plt.show(block=False)
plt.savefig(join(HOME, "test-elpy/figure.svg"))

```
## Notes
- `elpy-importmagic--add-import-read-args` now return a string instead of a list of one string (I didn't find any reasons to keep a list of one element).
- Importmagic do not seems to be able to clean modules imported with aliases...
- `elpy-importmagic-fixup`  try to get an import statement for variables if they are not defined.

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -2245,7 +2245,7 @@ prefix argument is given, prompt for a symbol from the user."
 ;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Import manipulation
 
-(defun elpy-importmagic--add-import-read-args (&optional object prompt)
+(defun elpy-importmagic--add-import-read-args (&optional object prompt ask-for-alias)
   (let* ((default-object (save-excursion
                            (let ((bounds (with-syntax-table python-dotty-syntax-table
                                            (bounds-of-thing-at-point 'symbol))))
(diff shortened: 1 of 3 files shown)


jorgenschaefer:
This looks very good, thank you very much for your contribution and work!

Could you squash the commits? I'll be happy to merge. :-)

muahah:
Done,
do you think adding import with alias deserve a new binding ?
I use it quite often, but I may be the only one.

jorgenschaefer:
I think the prefix argument is a good solution here. If people use this a lot, they can easily add a key binding themselves. Using an alias should be an unusual thing outside of the scipy/numby/matplotlib world.

The only thing I can imagine is some kind of configuration list that maps names to aliases for a fixed set of modules.

muahah:
I can make the prefix argument thing works by using `(interactive "i\nP")` in `elpy-importmagic-add-import`.
Is this the correct way of doing it ?

jorgenschaefer:
Yeah, that would work.

muahah:
Does it worth the trouble to make a new pull request ?

jorgenschaefer:
I'd accept it :-)

PR #1037, opened 2016-12-05
Add a suggestion for org-mode's ob-ipython package

yuvallanger:


jorgenschaefer:
Thank you for the contribution! I'm afraid I do not think it is useful for Elpy to reference every Python-related package out there, so I would prefer to do without this.

Issue #988, opened 2016-09-20
Indentation Problems


amirteymuri:
I have 2 difficulties with indentation matters in elpy:

1. In the example below when i choose one of the function definitions with `M-h` as a block and try to evaluate **only** that function with `C-c C-c` i get the TabError:

    >>> Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
      File "/tmp/py2585eSP", line 5
        if x % 2 == 0:
                     ^
    TabError: inconsistent use of tabs and spaces in indentation

2. In the code below when i am on the `if __name__ == "__main__:"` line and hit the Enter to go to the next line, the `if __name__ == "__main__:"` line gets also indented as seen below:

def genFn():
    x = 0
    while True:
        if x % 2 == 0:
            print("in IF")
            yield x
            print("after yield")
        if x > 10:
            return x    
        print("outside IF : %d" % x)
        x += 1
        print("outside IF : %d" % x)

def infinitFibs():
    a, b = 0, 1
    while True:
        a, b = b, a + b
        yield a

        if __name__ == "__main__":

Is there a way stopping these behaviors?

Greetings, Amir

jorgenschaefer:
Hello, and thanks for the report! Do these problems exist with Elpy disabled (and using `M-x python-shell-send-region` instead of `C-c C-c`)? If so, could you file a report with Emacs directly?

EmperorDali:
I am also having the TabError bug. It does *not* occur with `python-shell-send-region`.

amirteymuri:
With elpy disabled and using `python-shell-send-region` the Error does not happen. Besides i get a strange `IndentationError` in this another example when i try to evaluate the `def` using `C-M-x`:

    def foo(*args, **kwargs):
        print args, kwargs

This has raised this error **with elpy disabled**:

    >>> Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
      File "/home/amir/programming/python/tst.py", line 2
        if True:
               ^
    IndentationError: expected an indented block

and **with elpy enabled**:

    >>> Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
      File "/home/amir/programming/python/tst.py", line 2
        if True:
               ^
    SyntaxError: unexpected EOF while parsing

jorgenschaefer:
I have no idea why this behavior would change with Elpy enabled, sorry.

muahah:
I also have the `TabError`.
[...]

PR #1039, opened 2016-12-11
Fix apparition of tabs while removing indentation

muahah:
For some user (see #988), wild tabs seem to appear while sending code parts to python shell. 
This seems to be due (for me) to `indent-rigdly` in `elpy-shell--region-without-indentation`.
Using `untabify` after `indent-rigidly` fix the problem, but I don't think `indent-rigidly` is supposed to add tabs, so this just looks like a workaround...

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -1735,6 +1735,9 @@ If ASK-FOR-EACH-ONE is non-nil, ask before killing each python process.
         (indent-rigidly (point-min)
                         (point-max)
                         (- indent-level))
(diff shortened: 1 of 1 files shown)


jorgenschaefer:
Nice debugging work, and thank you for the PR! Could you add a comment to that line explaining why it is there? (Basically, something like "there seems to be a bug in indent-rigidly which introduces tabs", and if you have, with a link to the upstream bug report)

muahah:
Added.
I keep digging for this bug,
I cannot reproduce it on other modes, so it may still be related with python mode.

jorgenschaefer:
Thank you very much!

PR #1042, opened 2016-12-16
Add django shells

gopar:
Can load up a python shell that's already loaded with Django
environment. Can also set what shell they want to use. The
"--plain" flag is used to avoid formating errors. For example if we set
the shell to "bpython", it will not work. Using that flag makes shell
process use default interpreter.

Not sure if this deserves a keybiding. I didn't give it one.

### DIFF
--- a/elpy-django.el
+++ b/elpy-django.el
@@ -153,12 +153,16 @@ it will prompt for other flags/arguments to run."
 prompt for additional args"
   (interactive "P")
   ;; Due to formatting, "--plain" argument is you're safest bet
-  (let ((proj-root (file-name-base (directory-file-name (elpy-project-root))))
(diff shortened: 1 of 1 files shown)


gopar:
Hmm, looks like I can't see what the merge conflicts are. Any who, any updates on this? Thanks

jorgenschaefer:
Oops, this must have fallen through the cracks during my vacation in December, sorry! Looks good to me, could you squash the two commits (and fix that conflict)? I'll merge. Thanks :-)

gopar:
Looks like I might have to close this and reopen since I can't checkout out this pr to fix the merge issues. http://stackoverflow.com/q/36071272/2178312

gopar:
PR moved to  #1154

PR #1048, opened 2016-12-21
Add the possibility to use prefix arg in elpy-importmagic-add-import

muahah:
With a prefix arg, `elpy-importmagic-add-import` ask for an import alias.


muahah:
Following discussion in #1033

jorgenschaefer:

PR #1053, opened 2016-12-27
Fix compilation warnings

AdamNiederer:
Emacs complains when byte-compiling this file
```
elpy-django.el:126:37:Warning: ‘beginning-of-buffer’ is for interactive use
    only; use ‘(goto-char (point-min))’ instead.
elpy-django.el:117:87:Warning: ‘beginning-of-buffer’ is for interactive use
    only; use ‘(goto-char (point-min))’ instead.
```
This PR should fix it.

### DIFF
--- a/elpy-django.el
+++ b/elpy-django.el
@@ -113,10 +113,10 @@ require arguments in order for it to work."
           (with-temp-buffer
             (progn
               (insert help-output)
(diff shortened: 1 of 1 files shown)


jorgenschaefer:
Excellent, thank you! :-)

Issue #1073, opened 2017-02-07
Warn when disabling completion due to buffer size


teoric:
Would it be possible to emit a warning if the buffer ist so large that completion does not work?  It took me surprisingly long to find the reason why `elpy` worked in some files but not others. It would have been enough to get something like `Buffer larger than elpy-rpc-ignored-buffer-size -- disabling completion`.  

The variable does not appear to be [in the docs](http://elpy.readthedocs.io/en/latest/), either, it's only in the NEWS.

(Cannot have a look right now, but might look into a pull request if you don't beat me to it.)

gopar:
Should this issue be closed?  PR is already merged.

jorgenschaefer:
Thanks for noticing :-)

PR #1074, opened 2017-02-08
Warn if disabling completion

teoric:
Disabling completion can confuse users when happening silently. Addresses #1073 ("Warn" part).

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -3400,10 +3400,10 @@ If you need your modeline, you can set the variable `elpy-remove-modeline-lighte
      (company-mode 1)
      (when (> (buffer-size) elpy-rpc-ignored-buffer-size)
        (message
-	(concat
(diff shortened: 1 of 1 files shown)


jorgenschaefer:
Thank you for the contribution! Do you really need `(concat (format ...) ...)`? If you do this to break long lines, use `(format (concat ...) ...)`, but in general, you can just leave the format string as a long line.

jorgenschaefer:
(Those build errors don't seem to be related to your changes. Weird those.)

teoric:
Thank **you** for elpy! ;-)

Personally, I do not really like long lines, but feel free to change this.  I adapted the order of `format`/`concat` to your taste.  I had assumed that format should be applied to the shortest possible string, but it certainly cannot make much of a difference, and may be clearer in the other order.

jorgenschaefer:
Excellent! Thank you very much :-)

PR #1076, opened 2017-02-12
Display case-insensitive company candidates correctly

brotzeitmacher:
The problem was, that the candidate names were concatenated with "prefix" and "suffix". The prefix of a case-insensitive candidate includes the wrong case, so company did not display the actual name correctly.


jorgenschaefer:
Nice catch! Thank you!

Issue #544, opened 2015-04-24
Improved M-. / M-, functionality


jorgenschaefer:
`M-.` should jump to the most likely definition of the current symbol, and put the old position into the global mark ring.

`C-u M-.` should list all possible definitions, assignments, and also _uses_ of the thing at point, with the respective line from their respective file, in an compilation-like buffer, with the user being able to just jump to any of them as needed.

This issue replaces:
#401 Command to list all usages of the symbol at point
#476 New `M-.`/`M-,` API in Emacs 25

johnchanwork:
Nice feature. I've been waiting for this for awhile. And this is THE feature I missed switching from anaconda-mode.

PR #1077, opened 2017-02-12
Add the possibility to display usages for thing at point

brotzeitmacher:
Got the idea from Improved M-. / M-, functionality #544

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -2612,6 +2612,90 @@ Also, switch to that buffer."
         (select-window window)
       (switch-to-buffer "*Occur*"))))
 
+;;;;;;;;;;;;;;;;;;;;;
+;;; Find Usages
+
+(define-derived-mode elpy-view-usages-mode special-mode "elpy-usages"
+  (toggle-truncate-lines)
+  (setq next-error-function #'anaconda-mode-next-definition))
+
+(define-button-type 'elpy--usage-button
+  'action #'elpy-toggle-usage
+  'face nil)
+
+(defun elpy-insert-usage-button (name file offset)
+  "Insert button with NAME and save FILE and OFFSET."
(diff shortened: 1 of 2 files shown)


TauPan:
Uhm, may I ask why this was apparently never merged? I notice that travis-ci is failing, so maybe that was the reason? Would be sweet to have this in elpy ;-}

I expected this to be in 1.19 because of @jorgenschaefer 's adding this to the 1.19 milestone, but it's not in the released code.

brotzeit:
AFAIK elpy now works with `xref`. I'm not sure, but I think there was a PR some time ago. Take a look at `xref-find-references`.

galaunay:
Xref commands should indeed work with Elpy.

`xref-find-references` can be quite slow sometimes, If you are just interested by the usages for the thing at point, you may want to use `(xref-find-references (elpy-xref--identifier-at-point))` instead.

TauPan:
Hm, I guess I should re-read the documentation now and then ;) Thanks @brotzeit and @galaunay !

However this just returns references in the current buffer for me while e.g. elpy-rgrep-symbol returns quite a lot more (real references to the symbol in other python files, not just mentions of the string).

brotzeit:
That doesn't sound very helpful indeed. I haven't used python for a while, but I know that emacs lsp support keeps getting better. Maybe you should give https://github.com/emacs-lsp/lsp-python a try. But I don't know the state of the python lsp server. I've only tried it with rust.

galaunay:
The xref implementation uses jedi, so it should get references in other files as well (it does for me).

Python Language Server apparently uses jedi as well for references, so I would not expect fundamentally different results.

However, Xref implementation is new and has been a little bit chaotic (my bad).
There is certainly room for improvement.

PR #1078, opened 2017-02-12
Add commands to profile python scripts

muahah:
This pull request adds the function `elpy-profile-buffer-or-region` that runs the python profiler (`cProfile`) asynchronously on the current buffer (or active region) and displays the result with the wanted visualizer (snakeviz by default).


elpy.el
```diff
@@ -1737,6 +1737,84 @@ If ASK-FOR-EACH-ONE is non-nil, ask before killing each python process.
...
+If DONT-DISPLAY is non nil, don't display the profile results."
+  ;; clean log buffers
+  (condition-case nil
+      (with-current-buffer " *elpy-profile-log*"
+	(kill-this-buffer))
+    (error nil))
+  (condition-case nil
+      (with-current-buffer "*elpy-profile-log*"
+	(kill-this-buffer))
+    (error nil))
```
jorgenschaefer: `(ignore-errors (kill-buffer NAME))` is a lot shorter.

elpy.el
```diff
@@ -1737,6 +1737,84 @@ If ASK-FOR-EACH-ONE is non-nil, ask before killing each python process.
...
+  (let* ((prof-file (if in-dir (concat (file-name-sans-extension file) ".profile")
+		      (concat (make-temp-file "elpy-profile-" nil ".profile"))))
```
jorgenschaefer: Please trust Emacs to indent your code correctly. Also, add a newline after the condition to make this more readable.

elpy.el
```diff
@@ -1737,6 +1737,84 @@ If ASK-FOR-EACH-ONE is non-nil, ask before killing each python process.
...
+  "Elpy profile sentinel."
+  (let ((filename (file-name-nondirectory (process-get process 'file)))
+	(prof-file (process-get process 'prof-file))
+        (dont-display (process-get process 'dont-display)))
+    (if (not (string-equal string "finished\n"))
+	(progn
+	  (message  "[%s] Profiling failed" filename)
+	  (switch-to-buffer-other-window " *elpy-profile-log*")
+	  (read-only-mode)
+	  (rename-buffer "*elpy-profile-log*"))
```
jorgenschaefer: You can just `(rename-buffer OLD NEW)`. Though in general, I would recommend to just use the visible buffer name all the time – no need to hide this.
Also, switching the user's input away without any action of theirs is somewhat rude. Just use `display-buffer` to show it.
You can use `view-mode` to make it nicely scrollable using SPC etc.

gopar: `rename-buffer(OLD NEW)`  isn't valid. Instead he can do
```lisp
(with-current-buffer BUFF-NAME
      (rename-buffer NEW-NAME))
```

elpy.el
muahah: I am not sure to see what's wrong with the indentation here.
I re-indented everything to be sure.

elpy.el
muahah: Definitely, I made the change.

elpy.el
muahah: I let the buffer visible as suggested, and replace `read-only-mode` by `view-mode`

muahah:
With some more time, I can try adding some tests.

muahah:
Tests seem to work, but I'm not sure about the implementation.
I also added documentation for `elpy-profile-buffer-or-region`.

gopar:
I think it would be a good idea if the docs mentioned that the user needs something like `snakeviz` to be installed as well.

muahah:
I added a mention of snakeviz in the doc (thanks @gopar).
The second commit allows to set command with arguments in `elpy-profile-visualizer` (`(setq elpy-profile-visualizer "snakeviz -H 127.0.0.1 -p 8080")` for example)

jorgenschaefer:
Looks good to me. If you are happy with this as is, could you squash the commits and let me know? I'd be happy to merge. :-)

muahah:
I tried to use profiling as much as possible this last days.
I didn't run into any troubles so far so it should be good.

PR #1079, opened 2017-02-12
Add functions to send code parts to python shell.

galaunay:
I wrote those little functions that I use a lot to send code parts to shell:
`elpy-shell-send-def`
`elpy-shell-send-class`
`elpy-shell-send-paragraph`
I hope it can be of use for some other users.

`elpy-shell-send-def` and `elpy-shell-send-class` are smart enough to take into account the current line indentation to find the right function/class definition to send.
For example, calling `elpy-shell-send-def` on this situation will send `bar` definition to shell only when the current line is in its definition.
```Python
def foo(a, b):        # send foo
    def bar(a):       # send bar
        return a**2   # send bar
    b = b % a         # send foo
    if b > a:         # send foo
        return b      # send foo
    return a          # send foo
```

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -1631,6 +1631,105 @@ code is executed."
   (elpy-shell-display-buffer)
   (python-nav-forward-statement))
 
+(defun elpy--line-match-p (regexp)
+  "Return t if current line match REGEXP"
+  (save-excursion
+    (beginning-of-line)
+    (search-forward-regexp regexp (line-end-position) t)))
+
+(defcustom elpy-def-regex "[[:space:]]*\\_<\\(?:def\\|async[[:space:]]+def\\)\\_>[[:space:]]+\\(?9:[_[:alpha:]][_[:word:]]*(.*)\\):"
+  "Regular expression matching a function definition line.
+
+Group 9 is the function signature displayed in the minibuffer by `elpy-shell-send-def'."
+  :type 'regexp
+  :group 'elpy)
+
+(defcustom elpy-class-regex "[[:space:]]*\\_<\\(?:class\\)\\_>[[:space:]]+\\(?9:[_[:alpha:]][_[:word:]]*(.*)\\):"
+  "Regular expression matching a class definition line.
+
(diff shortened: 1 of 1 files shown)


jorgenschaefer:
Thank you for the contribution! But isn't this what `C-M-x` does?

muahah:
`C-M-x` always send the upper-most class or function definition (here, it always send `foo` definition).
It can be annoying in cases like this : 
```Python
if cond is True:          # send nothing  # `C-M-x` send if statement
    def foo(a, b):        # send foo      # `C-M-x` send foo
        def bar(a):       # send bar      # `C-M-x` send foo
            return a**2   # send bar      # `C-M-x` send foo
        b = b % a         # send foo      # `C-M-x` send foo
        if b > a:         # send foo      # `C-M-x` send foo
            return b      # send foo      # `C-M-x` send foo
        return a          # send foo      # `C-M-x` send foo
                          # send foo      # `C-M-x` send foo
    def foo2(a, b):       # send foo2     # `C-M-x` send foo
        return a + b      # send foo2     # `C-M-x` send foo
```
Explanation is straightforward when looking at the definition of `python-shell-send-defun`: it searches for function definition with level 0 indentation, and when it can't find one, it takes the first function definition (`foo` in this case).

jorgenschaefer:
That's some silly definition for `C-M-x` … I would have expected it to send the current top level block. Oh well.

I think your definitions are better suited in Emacs' `python.el`. Have you considered contributing them there?
If not, or if they are rejected, would you mind putting them in a separate Emacs Lisp file, instead of the main elpy.el? Something like `elpy-shell.el` I think would make sense.

Tests would be nice, too, but not mandatory :-)

muahah:
I will try proposing a modification of `python.el`, somebody may also have more information on why `C-M-x` behave like this.
I'll come back here when I have more information.
thanks for your advice.

rgemulla:
I guess this PR can be closed now. @galaunay Do you agree?

galaunay:
@rgemulla, yes, of course.
Closing...

PR #1080, opened 2017-02-13
Add the possibility to display usages for thing at point

brotzeitmacher:
Next try :D

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -2612,6 +2612,89 @@ Also, switch to that buffer."
         (select-window window)
       (switch-to-buffer "*Occur*"))))
 
+;;;;;;;;;;;;;;;;;;;;;
+;;; Find Usages
+
+(define-derived-mode elpy-view-usages-mode special-mode "elpy-usages"
+  (toggle-truncate-lines))
+
+(define-button-type 'elpy-usage-button
+  'action #'elpy-toggle-usage
+  'face nil)
+
+(defun elpy-insert-usage-button (name file offset)
+  "Insert button with NAME and save FILE and OFFSET."
+  (insert-text-button name
+                      'type 'elpy-usage-button
+                      'file file
+                      'offset offset))
+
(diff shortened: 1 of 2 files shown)


jorgenschaefer:

Please do adjust (and preferable, add) tests to your changed code.

Do you think it would make more sense to use Emacs 25's new `xref.el` facility to go to usages?

brotzeitmacher:
Thanks for the hint regarding xref.el. I agree that it would be better to use this library, however I don't know if I'll take the time to rewrite it soon, as my implementation seems to work.

Unfortunately I don't have any idea how to adjust/add tests so far. Elpy is the first project I'm contributing to, so this stuff is all fairly new to me and I'm glad, that I'm able to contribute anything at all. But I will add it as soon as I know how :)

jorgenschaefer:
In theory, you should be able to run `./scripts/setup` to set up the development environment, and then `./scripts/test` to run the tests. It's a good idea to get used to writing tests for the code you write, it's a sign of professionalism.

As for this specific PR, though I really like your work, I think it'd be best if Elpy would use xref. Maybe you can publish your code yourself as a separate package for others who want to use it? :-)

brotzeitmacher:
Thanks for the explanation. I'll take a look at it.

The stuff I've written depends on elpy's backend, so I guess there's no sense in making a seperate package. Maybe you can open an issue, that motivates others to take care of it. The lack of this feature was the reason why I used anaconda besides elpy.

jorgenschaefer:
> The stuff I've written depends on elpy's backend, so I guess there's no sense in making a seperate package.

You can write a package that depends on Elpy :-)

jorgenschaefer:
> The stuff I've written depends on elpy's backend, so I guess there's no sense in making a seperate package.

Ah right, makes sense of course. Hm. I'll make an issue out of it, good idea.

PR #1092, opened 2017-03-05
Use Xref library for completion, find definition and usage

galaunay:


muahah:
This is an attempt of using the xref library for definition and usages finding.

Currently, `xref-find-definition` and `xref-find-references` work.

Remaining work to be done : 
- Currently the symbol identifiers are their position in the current buffer, some xref functionnalities will only work if identifiers are strings (function or variable names).
- `xref-find-apropos` still need to be implemented.
- All of this is only implemented for jedi (no rope support).

Additionnaly, I tried to add `rpc-get-identifier` to `jedibackend.py` without success.
I apparently missed something...

brotzeitmacher:
anybody still working on this ?

muahah:
I should have some time soon to make a working version of it.
But all xref functionalities will not be available due to the way jedi works (see my comment above).
Question is : Is it better to have a partial implementation of xref or to stick to the good old elpy functions ?

muahah:
### Currently work : 
- `xref-find-definition`
- `xref-find-definition-other-window`
- `xref-find-definition-other-frame`
- `xref-pop-marker-stack`
- `xref-find-reference`
- Everything in `*xref*` buffer (resulting from `xref-find-reference`) except `xref-query-replace-in-results` (that does not work in elisp too for me...).
### Currently don't work : 
- `xref-find-apropos` : need to be able to retrieve all file identifiers, which seems complicated with jedi.

If somebody find a way to get identifiers from jedi, `xref-backend-identifier-completion-table` will need to be implemented too, to make identifier completion works.

muahah:
After reading the jedi API It was in fact not that hard to get the list of defined names...
All xref functionnalities should run with a little more work.

### Currently work : 
- `xref-find-definition`
- `xref-find-definition-other-window`
- `xref-find-definition-other-frame`
- `xref-pop-marker-stack`
- `xref-find-reference`
- Everything in `*xref*` buffer (resulting from `xref-find-reference`) except `xref-query-replace-in-results` (that does not work in elisp too for me...).

### Todo :
- Implement `xref-find-apropos` using `elpy-rpc-get-names` (already implemented and working)
- Add tests
- Add debugging capabilities to `rpc-get-names` in `jedi-backend.py`

elpy.el
```diff
@@ -3292,13 +3293,161 @@ Returns a possible multi-line docstring."
...
+            success error))
+
+;;;;;;;;;;;;;;;;
+;;; Xref backend
+
+;;;###autoload
+(defun elpy--xref-backend () 'elpy)
+
+(defun elpy-xref--set-backend ()
+  (setq-local xref-backend-functions 'elpy--xref-backend))
```
jorgenschaefer: Shouldn't this be a list? Is there a reason why we preclude other xref backends?

elpy.el
```diff
@@ -3292,13 +3293,161 @@ Returns a possible multi-line docstring."
...
+       for line = (with-current-buffer (find-file-noselect file)
+                    (save-excursion
+                      (goto-char (+ pos 1))
+                      (buffer-substring (line-beginning-position) (line-end-position))))
+       for summary = (format elpy-xref--format-references linenumber line)
+       for loc = (xref-make-elpy-location file (+ pos 1) line)
+       collect (xref-make summary loc)))))
+
+(defun xref-find-references ()
+  "Overwrite the current `xref-find-references` to avoid completion asking"
```
jorgenschaefer: Isn't this kind of rude, unconditionally redefining a function in another library? This affects non-Elpy uses of xref, too.

elpy/jedibackend.py
```diff
@@ -173,6 +173,29 @@ def rpc_get_usages(self, filename, source, offset):
 
         return result
 
+    def rpc_get_names(self, filename, source):
+        """Return the list of possible names"""
+        # Remove form feed characters, they confuse Jedi (jedi#424)
+        source = source.replace("\f", " ")
```
jorgenschaefer: This sounds like something more generic :-D

muahah:
I will take a look to this soon, test functions are on my todo list.
### Currently work :
- Everything, except `xref-query-replace-in-results` in `*xref*` buffers (that does not work in elisp too for me...).

### Todo :
- Add tests
- Add debugging capabilities to rpc-get-names in jedi-backend.py (?)
- Fix Travis error : Cannot open load file: xref (I imagine this is linked to Travis since all the tests pass on my computer)

elpy.el
muahah: Fixed. It was a temporary workaround.

elpy.el
muahah: You are right, It use now a `add-hook` in the mode definition.

elpy/jedibackend.py
muahah: I used "names" because jedi does, maybe "definitions" or "identifiers" will sound better ?

muahah:
Until now, I used buffer position as identifier (because it was easier for jedi to deal with), but some xref functionalities  (notably `xref-find-references`) need to have symbols as identifiers.
This new version use symbols as identifiers (and everything work fine).

### Still todo : 
- Add tests
- Implement `get_names` function for rope (else, `xref-find-apropos` and `xref-find-references` won't work)
- Remove redundant elpy functions (`elpy-goto-definition`, `elpy-goto-definition-other-window`, ...)
- Fix Travis error : Cannot open load file: xref (I imagine this is linked to Travis since all  tests pass on my computer)

muahah:
I added tests and fix some minor issues.
I am not sure about the tests implementation, so this may need some verifications.

Three questions remain : 
- How to fix the travis error...
- Should `elpy-goto-definition` and `elpy-goto-definition-other-window` be removed in favour of xref functions or not.
- Is it useful to implement `rpc_get_names` for rope.

jorgenschaefer:
> - How to fix the travis error...
> - Should `elpy-goto-definition` and `elpy-goto-definition-other-window` be removed in favour of xref functions or not.

For people stuck with older Emacsen, I'd keep them around. Elpy should support older Emacsen out of the box. (Rough rule of thumb, if Debian Stable's Emacs version should be supported)

> - Is it useful to implement `rpc_get_names` for rope.

If you want. I'd not bother with Rope anymore I think.

muahah:
I think we are finally good.

`elpy-xref` backend is now only loaded if xref is available (related tests are also ran only if xref is available).
As functions necessary to the `elpy-xref` backend are only defined for jedi, `elpy-xref` backend is now deactivated if the used rpc backend is not jedi.

Let me know if you see remaining changes to make, else, I'll squash the commits to make something presentable.

elpy.el
```diff
@@ -42,6 +42,8 @@
 (require 'ido)
 (require 'json)
 (require 'python)
+(when (featurep 'xref)
+  (require 'xref))
```
jorgenschaefer: You should use `(require 'xref nil t)` here.

elpy.el
```diff
@@ -537,6 +539,8 @@ virtualenv.
   :lighter " Elpy"
   (when (not (derived-mode-p 'python-mode))
     (error "Elpy only works with `python-mode'"))
+  (when (featurep 'xref)
+    (add-hook 'xref-backend-functions #'elpy--xref-backend nil t))
```
jorgenschaefer: Maybe use `(when (boundp 'xref-backend-functions) ...)` here? It's usually better to test for the functionality used instead of the feature.

elpy.el
```diff
@@ -3309,13 +3313,165 @@ Returns a possible multi-line docstring."
...
+  (elpy-rpc "get_names"
+            (list buffer-file-name
+                  (elpy-rpc--buffer-contents)
+                  (- (point)
+                     (point-min)))
+            success error))
+
+;;;;;;;;;;;;;;;;
+;;; Xref backend
+(when (featurep 'xref)
```
jorgenschaefer: You do not need to conditionally define these. As long as they are not used, they can happily be defined, I think?

elpy.el
jorgenschaefer: Also, I think this is missing the conditional definition of `M-.` and `M-,` to use the xref functionality … or am I missing something? I.e. if xref is available, use that instead of the old elpy functionality?

elpy.el
muahah: Done

elpy.el
muahah: I forgot to handle the keybindings...

`xref` commands should be bound by default (see [xref doc](https://www.gnu.org/software/emacs/manual/html_node/emacs/Looking-Up-Identifiers.html#Looking-Up-Identifiers)).
So we just need to define `M-.` and `M-,` to elpy functionalities if xref is not available.
The problem is that elpy do not use exactly the same shortcuts than xref (`elpy-goto-definition-other-window` is bound to `C-x 4 M-.`, but `xref-find-definition-other-window` is bound to `C-x 4 .`).
I added one additional shortcut to `xref-find-defintion-other-window` to avoid confusing people used to the elpy shortcuts.

I also add a paragraph for presenting the `xref` functionalities in the documentation.

elpy.el
muahah: Done

muahah:
Done, but I have trouble to pass the tests, as `cl-method` is said to be void.
I tried adding `cl-lib` to the dependencies, but it's apparently not enough.

jorgenschaefer:
I think you need `cl-generic` for `cl-defmethod`.

muahah:
It gives me a different error:
`Cannot open load file: no such file or directory, cl-generic`
Tests pass on my emacs, maybe something to do with travis ?

muahah:
I was obliged to add back the conditional definition of the xref functions.
Indeed, some of them use classes defined in xref (and consequently raise an error if xref is not loaded).

On another subject, is there any reason for Travis not to test the elpy package on emacs 25 ?

jorgenschaefer:
> I was obliged to add back the conditional definition of the xref functions.

Ah, sorry for leading you on the wrong path there :-(

> On another subject, is there any reason for Travis not to test the elpy package on emacs 25 ?

No. :-D

muahah:
Maybe modifying Travis tests deserves another pull request.
I think I am done with this, it does pass the tests on emacs 24 and works on emacs 25 (intensively tested on my side).
If it's good on your side I can make a last review and squash the commits.

jorgenschaefer:
lgtm! :-) I'll be happy to merge as soon as you give the go.

muahah:
I corrected some typo and re-ran the test once again.
Everything works fine.
It took the time but I think this is it !

galaunay:
I apparently closed this before you had time to merge it...
It was functional three month ago, I just rebased it on the current master branch.

jorgenschaefer:
Thank you very much for all the work you put in here, and your patience :-)

elpy-pkg.el
```diff
@@ -5,5 +5,5 @@
                   (highlight-indentation "0.5.0")
                   (pyvenv "1.3")
                   (yasnippet "0.8.0")
-                  (s "1.11.0"))
-                )
+                  (s "1.11.0")
+                  (cl-generic "1.0"))
```
suutari-ai: Does this really need the newest version of cl-generic? I can't seem to find the 1.0 version for Emacs 24.5.1 Only 0.3 is available here.

It would be much appreciated if Elpy would still work Emacs 24.5.d

PR #1100, opened 2017-03-16
Add project shell with project root in sys.path

### LINKED ISSUE #55: Per-project Python shells
It could be useful to have per-project shells, like there are per-file dedicated shells right now. Those shells also could get the project root added to sys.path.

gopar:
Looks like this closes #55 (Wow that's old \o/)

One thing to note is that `C-c C-z` does go back to the last previous py buffer but does not go back to proj shell when doing `C-c C-z` in py buffer(Due to default behaviour)

### DIFF
--- a/docs/ide.rst
+++ b/docs/ide.rst
@@ -145,7 +145,6 @@ Emacs can run a Python interpreter in a special buffer, making it much
 easier to send code snippets over.
 
 .. command:: elpy-shell-switch-to-shell
-   :kbd: C-c C-z
 
    Switch to buffer with a Python interpreter running, starting one if
    necessary.
@@ -154,6 +153,13 @@ easier to send code snippets over.
    This process will inherit the current virtual env. If you want to
    start a Python interpreter with a different virtual env, you can
    either kill the existing one, or rename the buffer.
+
+.. command:: elpy-shell-switch-to-project-shell
+   :kbd: C-c C-z
+
+   Switch to a Python interpreter buffer with the current project
+   already loaded into sys.path. Each project can have their own
+   Python interpreter running.
 
 .. command:: elpy-shell-send-region-or-buffer
    :kbd: C-c C-c


jorgenschaefer:
This looks great, thank you! Do you know of any particular reason why this behavior should not be the default? I.e. make `C-c C-z` always switch to a project-specific shell, and get rid of global/dedicated shells?

gopar:
I would much prefer this to be the default since I'd rather have a shell that already has my project loaded in.  I rarely used `C-c C-z` since it didn't have the project in sys.path

jorgenschaefer:
Let's make it so, then. I don't see a need for "dedicated shells" as such anymore, so let's get rid of that concept. Make this behavior the default.

Would you like to do the changes?

gopar:
I don't mind. I'm guessing for now only rebind `C-c C-z` to bring up the project shell and not remove any of the other shell functions? I'm assuming other people are depending on those functions. But ultimately the decision is yours. Let me know what you want me to do. \o/

jorgenschaefer:
Hm. Changing the default binding as a first step sounds like a good approach. We can remove the other functions slowly. Nice idea, thanks!

gopar:
Rebinded `C-c C-z` to use project shell and added it to the docs.

jorgenschaefer:
Hm. If you make `elpy-shell-switch-to-shell` do a project shell (which I approve!), then there's no need to add another function to the documentation, is there? And at that point, you can just change `elpy-shell-get-or-create-process` and get rid of the whole elpy-dedicated-shells things, as we lost backwards-compatibility anyhow. (I'm fine with all of this :-D) Or am I missing something?

gopar:
Oh, that's what you want :P 

I was afraid of changing the existing `elpy-shell-*` functions so I created a new one. Alrighty, I'll do the fixes then \o/

galaunay:
@gopar could you check if #1232 with option **project-root** is behaving like your implementation here ?

gopar:
@galaunay looks like it is (Haven't checked it) but it has the line that should make it behave so. Will test it out later. Also, if that PR is accepted feel free to remind me to close this one \o/

rgemulla:
This is mostly addressed by #1209, except that the sys.path is not appended. @galaunay I sugget to add the line below after the shell is started for the project.

```
(python-shell-send-string (format "import sys;sys.path.append('%s')" (elpy-project-root))))))
```

gopar:
@galaunay Can you add the appropriate line so that I can close this PR? Thanks

galaunay:
I have not forgotten about this :)
I planned to add it with #1232, but I think this is a better idea to split the PR.
I made and merged #1235.
I close this PR.

PR #1118, opened 2017-04-11
Add elpy-test-django-with-manage check when there's no module found for Django test runner.

zhangyiant:


jorgenschaefer:
Thank you very much!

PR #1119, opened 2017-04-11
Fix typo

lgp171188:
'environment' was wrongly spelt as 'environnement'


jorgenschaefer:
Nice catch, and thank you for fixing it! :-)

PR #1125, opened 2017-04-21
Fix no-name buffer error when using `elpy-shell-kill-all`

muahah:
`elpy-shell-kill-all` run into an error  if a buffer have no name (`(buffer-name buffer)` returning `nil`).
This commit just add a preliminary check to ensure we work only with named buffer.
This bug is not easy to reproduce, so I don't know exactly what kind of buffer return no name...


jorgenschaefer:
Interesting! Thank you :-)

A killed buffer does not have a name anymore, i.e. `(let ((b (generate-new-buffer "FOO"))) (kill-buffer b) (buffer-name b))`, but then it should not be returned by the loop. Weird.

Issue #1147, opened 2017-06-25
elpy-shell-kill-all does not find the *Python* buffer


bgoodr:
Hi,

I think I have found a bug and a possible fix:

elpy-shell-kill-all is bound to C-c C-k.  I expected it to kill the
`*Python*` buffer but it does not find it. Looking at the copy of elpy
I have (elpy-20170613.1339) shows:

    (defun elpy-shell-kill-all (&optional kill-buffers ask-for-each-one)
      "Kill all active python shells.
    
    If KILL-BUFFERS is non-nil, also kill the associated buffers.
    If ASK-FOR-EACH-ONE is non-nil, ask before killing each python process.
    "
      (interactive)
      (let ((python-buffer-list ()))
        ;; Get active python shell buffers and kill inactive ones (if asked)
        (loop for buffer being the buffers do
              (when (and (buffer-name buffer)
                         (string-match "^\*Python\\\[.*\\]\*$" (buffer-name buffer))
                         )
                (if (get-buffer-process buffer)
                    (push buffer python-buffer-list)
                  (when kill-buffers
                    (kill-buffer buffer)))))
        (cond
         ;; Ask for each buffers and kill
         ((and python-buffer-list ask-for-each-one)
          (loop for buffer in python-buffer-list do
                (when (y-or-n-p (format "Kill %s ?" buffer))
                  (delete-process buffer)
                  (when kill-buffers
                    (kill-buffer buffer)))))
         ;; Ask and kill every buffers
         (python-buffer-list
          (if (y-or-n-p (format "Kill %s python shells ?" (length python-buffer-list)))
              (loop for buffer in python-buffer-list do
                    (delete-process buffer)
                    (when kill-buffers
                      (kill-buffer buffer)))))
         ;; No shell to close
         (t
          (message "No python shell to close")))))

The expression:

    (string-match "^\*Python\\\[.*\\]\*$" (buffer-name buffer))

Does not seem to match the buffer named `*Python*`.

Here is my fixed version that uses the handy rx macro:

    (defun elpy-shell-kill-all (&optional kill-buffers ask-for-each-one)
      "Kill all active python shells.
    
    If KILL-BUFFERS is non-nil, also kill the associated buffers.
    If ASK-FOR-EACH-ONE is non-nil, ask before killing each python process.
    "
      (interactive)
      (let ((python-buffer-list ()))
        ;; Get active python shell buffers and kill inactive ones (if asked)
        (loop for buffer being the buffers do
[...]

PR #1149, opened 2017-07-01
Fix bug in elpy-shell-kill-all where it does not find the *Python* bu…

bgoodr:
…ffer(s)

This is a fix for the problem reported in
https://github.com/jorgenschaefer/elpy/issues/1147#issue-238358628

This replaces the original regular-expression for one that does not
mandate that the buffer name contains a square-bracketed
expression. It is unknown why it needed to match the square-bracketed
expression in the first place, since the only buffer names I have seen
in practice being created by elpy are "*Python*". Perhaps in the past
there were allowed to be multiple interactive Python buffers?
Regardless, make it match either with or without the square-bracketed
form of buffer names.


jorgenschaefer:
Thank you very much for your work! :-)

PR #1154, opened 2017-07-12
Add Django Shells

gopar:
Reference to #1042 since I could no longer update the branch (I deleted) I had to create a new pr with same changes. Let me know if there's anything you'd like me to change \o/


gopar:
Essentially yes. But from my understanding, it does some other stuff to setup the environment in the shell as well.

jorgenschaefer:
> But from my understanding, it does some other stuff to setup the environment in the shell as well.

Interesting. What other stuff? :-)

gopar:
Nvm, I took a quick look and looks like nothing else happens. Closing pr

PR #1175, opened 2017-09-14
cl-generic dependency typo fixup

ufnalbartosz:
Fix due to dependency error while installing elpy.


jorgenschaefer:
Thank you for this contribution! I actually think this works better without the dependency, even :-)

PR #1180, opened 2017-09-23
elpy.el: fix typo __main__ -> __name__

georgek:
Sorry for the tiny pull requests...


jorgenschaefer:
Oops, good catch! :blush: And tiny pull requests are the best! Thank you very much!

Issue #1182, opened 2017-09-28
Improvements for use with interactive Python


rgemulla:
I created a Spacemacs layer that integrates ```elpy``` with Spacemacs. The layer allows to use ```elpy``` in a style similar to [ESS](https://ess.r-project.org/) and extends some of its capabilities. Additions include more flexible evaluation commands, echoing of inputs sent to the Python shell in the shell buffer, echoing of Python shell outputs in the echo area, flashing of statements being evaluated, as well as a few minor bugfixes/improvements of ```elpy``` and ```python``` packages. See [here](https://github.com/rgemulla/spacemacs-layers/tree/master/+lang/elpy) for a more detailed description and source code.

Almost all of this functionality is not Spacemacs-specific, and I feel it may be useful for ```elpy``` users in general. I am opening this issue to gauge whether there is interest to directly integrate some of these additions into ```elpy```.

jorgenschaefer:
Hello, and thank you for all your work! It is greatly appreciated :-)

Generally, I'm quite interested in new features, given they are useful for many people and they do not add too much of a maintainability burden … small helper functions are great. As Python's use case between (less-interactive) application development and (more-interactive) numpy/scipy use differs quite a bit, do you think it would make sense to have a separate minor mode for the more-interactive functions? I do not know if there's enough room for them to get a default key binding otherwise.

rgemulla:
If you feel that there is enough interest (or have enough interest yourself), I'd be happy to help integrating. It shouldn't be that much work.

I fully agree that application developers won't care much about these additions (and wouldn't even notice them). They are tailored for interactive use and (in my eyes) allow for much higher editing and evaluation efficiency than by, say, using Jupyter notebooks.

The "echoing" additions go well with elpy and do not require new keybindings. I already created custom variables in the elpy group that control those features:

```lisp
(defcustom elpy/shell-display-buffer-after-send nil
  "Whether to show the Python shell after sending something to it."
  :type 'boolean
  :group 'elpy)

(defcustom elpy/shell-echo-output 'when-shell-not-visible
  "Whether to echo the Python shell output in the echo area after input has been sent to the shell.

  Possible choices are nil (=never), when-shell-not-visible, or
  t (=always)."
  :type '(choice (const :tag "Never" nil)
                 (const :tag "When shell not visible" when-shell-not-visible)
                 (const :tag "Always" t))
  :group 'elpy)

(defcustom elpy/shell-echo-input t
[...]

PR #1185, opened 2017-09-29
Improvements for interactive Python

rgemulla:
Implements the additions of #1182


rgemulla:
Might be worth to integrate #1079 as well. These two pull requests conflict.

rgemulla:
There are 6 test failures:

```
6 unexpected results:
   FAILED  elpy-shell-get-or-create-process-should-return-process
   FAILED  elpy-shell-send-region-or-buffer-should-display-but-not-select-buffer
   FAILED  elpy-shell-send-region-or-buffer-should-notify-of-removing-main
   FAILED  elpy-shell-send-region-or-buffer-should-send-buffer-without-region
   FAILED  elpy-shell-send-region-or-buffer-should-send-region-if-active
   FAILED  elpy-shell-switch-to-shell-should-switch-to-shell-buffer
```
Not sure why those fail. I tried these things, and they seem to do what's said behind "should". Help is appreciated.

rgemulla:
I added an option to send the current function definition to the shell as in #1079. 


Here is a summary of the current behavior. One may also build tests around this.
```python
# Expected behaviour of elpy-shell-send-*-and-step functions, where * is either
# statement (s), top-level statement (ts), function (f), or group (g). The
# original line the point is on is indicated by ###. Lines marked X should
# be sent, and the point should move to the beginning of the statement on the
# line marked P. E means that an error message should be shown and point not moved.
#
# Note that the behaviour of the elpy-shell-send-*-and-step functions indicated
# below assumes that the end-of-line-comments are not present.

# Example
print("Statement 1")

print("Statement 2")
def f1 ():
    def f11():
        print("f11")
    if True:
        print("yes")
    else:
        print("no")
# comment 1
print("Statement 3")

print("Statement 4")

# Expected behavior 1
                           ## | s | ts | f | g
print("Statement 1")       ## |   |    |   |
                           ## |   |    |   |
print("Statement 2")       ## |   |    |   | X
def f1 ():                 ## |   | X  | X | X
    def f11():             ## |   | X  | X | X
        print("f11")       ## |   | X  | X | X
    if True:               ## | X | X  | X | X
        print("yes")       ## | X | X  | X | X
    else: ###              ## | X | X  | X | X
        print("no")        ## | X | X  | X | X
# comment 1                ## |   |    |   | X
print("Statement 3")       ## | P | P  | P | X
                           ## |   |    |   |
print("Statement 4")       ## |   |    |   | P


# Expected behavior 2
                           ## | s | ts | f | g
print("Statement 1")       ## |   |    |   |
                           ## |   |    |   |
print("Statement 2")       ## |   |    |   | X
def f1 ():                 ## |   | X  |   | X
    def f11():             ## |   | X  | X | X
        print("f11") ###   ## | X | X  | X | X
    if True:               ## | P | X  | P | X
        print("yes")       ## |   | X  |   | X
    else:                  ## |   | X  |   | X
        print("no")        ## |   | X  |   | X
# comment 1                ## |   |    |   | X
print("Statement 3")       ## |   | P  |   | X
                           ## |   |    |   |
print("Statement 4")       ## |   |    |   | P


# Expected behavior 3
                           ## | s | ts | f | g
print("Statement 1")       ## |   |    |   |
                           ## |   |    |   |
print("Statement 2")       ## |   |    |   | X
def f1 ():                 ## |   |    |   | X
    def f11():             ## |   |    |   | X
        print("f11")       ## |   |    |   | X
    if True:               ## |   |    |   | X
        print("yes")       ## |   |    |   | X
    else:                  ## |   |    |   | X
        print("no")        ## |   |    |   | X
# comment 1 ###            ## |   |    | E | X
print("Statement 3")       ## | X | X  |   | X
                           ## |   |    |   |
print("Statement 4")       ## | P | P  |   | P
```

@galaunay Is this what you had in mind? Is what is called "group" here what you intended by paragraph? Next step planned is to also echo the function name, like in your code, and to do the same for class definitions.

galaunay:
That was exactly what I meant by paragraph.
I think I can close my own pull request, feel free to reuse any part of my code if you need it.
I will follow  this with interest, thank you for your work.

rgemulla:
Thanks galaunay, I'll do that!

rgemulla:
I added the missing features from the pull request #1079 of @galaunay. 

I also added keybindings to all the send-to-shell commands. They are not necessarily convenient, but they should be more or less intuitive and they don't clash with any other bindings.

@jorgenschaefer: I am now happy with features and code style. The test failures mentioned above still occur, but unfortunately I can't investigate right now because I can't get cask working on my Windows machine. As mentioned above, the functions do what they should do, though, at least on my machines. My current guess is that we need to fix the tests somehow.

elpy.el
```diff
@@ -4166,3 +3925,8 @@ which we're looking."
 
 (provide 'elpy)
 ;;; elpy.el ends here
+
+;;;### (autoloads nil nil ("elpy-django.el" "elpy-pkg.el" "elpy-profile.el"
+;;;;;;  "elpy-refactor.el") (22990 26173 0 0))
```
jorgenschaefer: This looks weird – probably should not be in this file at all.

elpy-shell.el
```diff
@@ -0,0 +1,976 @@
+;;; elpy-shell.el --- Interactive Python support for elpy -*- lexical-binding: t -*-
```
jorgenschaefer: Thank you for making this a new file, Elpy sorely needs more modularization :-)

elpy.el
```diff
@@ -444,6 +433,36 @@ option is `pdb'."
 
     (define-key map (kbd "M-TAB") 'elpy-company-backend)
 
+    ;; key bindings for elpy-shell
+    (define-key map (kbd "C-c e e") 'elpy-shell-send-statement)
```
jorgenschaefer: `C-c <char>` is a reserved key for the user and should not be used by modes. We could bind something to `C-u C-RET` or `C-u C-c C-c` for this, maybe. Or you could use `C-c C-y` as a prefix if you like (sending it to the other buffer is a bit like yanking it there maybe).

Thinking about stealing `C-c RET` back from importmagic, too …

rgemulla:
I like C-c C-y and updated the bindings accordingly. After prefix C-c C-y, ctrl is used for "and-step" and shift for "and-go". This should be reasonably efficient to type, although I'd expect heavy users of this functionality to rebind it to other keys themselves.

galaunay:
I took a look to the Travis logs, apparently the python failed tests are gone now.
Travis logs only show this error:
```
Symbol’s function definition is void: string-remove-suffix
```
And indeed to use this function you will need to add `(require 'subr-x)`.
Problem is that `subr-x` has been added with emacs-24.4, and this will make emacs 24.3 no longer supported by elpy...
Maybe a better idea would be to replace functions from `subr-x` by functions from `s` (`string-remove-suffix` => `s-chop-suffix`, `string-trim` => `s-trim`, `string-empty-p` => `s-blank?`, ....), as elpy already depends on `s`.

I ran the full test suite with these replacements on your branch and it work.

rgemulla:
Good catch, thanks! Is it possible to my branch (so that we can merge it in there first) or send a patch?

galaunay:
Here's the patch :
[0001-Replace-subr-x-functions-with-s-ones.patch.txt](https://github.com/jorgenschaefer/elpy/files/1399565/0001-Replace-subr-x-functions-with-s-ones.patch.txt)

Hope you will be able to merge your work soon :)

rgemulla:
Thanks a lot!

jorgenschaefer:
I suspect that would mean adding `s` as a requirement? Not sure that's worth it.

At the end of `elpy.el` is a section called "Backwards compatibility" – you can conditionally define functions missing in older Emacsen there, with some basic implementation. It does not have to be efficient or fast. Once Elpy moves to newer Emacsen as minimum requirement, this makes it easy to get rid of the workarounds.

jorgenschaefer:
Otherwise this looks good to me!

rgemulla:
@galaunay mentioned that ```elpy``` already depends on ```s```. I didn't check, but if that's true, then his patch (which is already integrated) fixes any backward compatibility issues and does not add new dependencies.

galaunay:
I though about `s` because it is in the requirement in the cask file (was apparently added in 2016-08-24 for django).

jorgenschaefer:
Ah yes, then you need to add the `(require 's)` thing to `elpy.el` :-)

rgemulla:
Done.

rgemulla:
There are more compatibility issues. I fixed one (with ```python-info-encoding```), but now the advices via ```advice-add``` fail (introduced in Emacs 24.4 afaik). 

Before going along, I'd like to hear your opinion on how to proceed. I see two options (other than dropping Emacs <=24.3 support, of course): 

(1) Revert back to the old advice system.
(2) Disable the advices on older Emacs versions. 

For option 1, I feel that it's kind of weird to still do this, although it shouldn't be a big deal to implement. 

For option 2, disabling the advices practically means that the shell echoing feature would only work on Emacs 24.4 and above.

jorgenschaefer:
I just checked – Debian stable ships with 24.5, oldstable with 24.4. I actually think we can drop the 24.3 dependency!

rgemulla:
Sounds good to me. 

I'll revert the last three commits once I get your go; no need to have these changes/dependencies then.

jorgenschaefer:
Sure! Make sure to remove the entries in the travis config and add a suitable dependencies to elpy-pkg.el and elpy.el, though :-)

rgemulla:
Ok, I bumped up the required Emacs version to 24.4. There is quite some code in elpy.el that can go now, but I did not remove it (it doesn't belong to this pull request).

I also changed elpy-shell.el so that it does not overwrite the ```python-shell-send-file``` function of python.el anymore (which is was a dirty hack). There are no more TODOs from my side (assuming that all test are passed clearnly now).

jorgenschaefer:
Very nice. Does this mean this branch is ready for merging? Are you happy with the number of commits and how they are split up?

rgemulla:
Yes, each commit only addresses a single issue. Also, all tests go through now. I'd say, it's ready to go.

I do want to add one more feature in the future: execute "codecells" of exported ipython/jupyter notebooks as well as functions to insert codecells and to convert ipython files to ipython notebooks (i.e., the other way around). Although all of this is implemented in my Emacs customization code already, I think this can go to a separate pull request later. It's probably useful only to a small set of people (but it's very useful to me).

jorgenschaefer:
Thank you very much for your excellent work!

rgemulla:
Thank you for your support!

Let me know if I can help to provide documentation for the new features (and where to do it).

jorgenschaefer:
Documentation would be nice :-) Something somewhere on the wiki at least, though ideally, adding the functionality to the documentation in `docs/` would be nice!

rgemulla:
Sure! It may take a small while, I'll open another PR once done.

PR #1194, opened 2017-10-16
elpy-module-company: no need for `format' inside `message'

sam-s:
minor code simplification and message improvement


jorgenschaefer:
Good catch, but I think this PR has a few commits that should not be there?

sam-s:
no idea how that happened.
should have been just one patch.

jorgenschaefer:
Nope, the branch has conflicts :-(

PR #1207, opened 2017-11-03
elpy-rgrep-symbol: allow editing the symbol at point with C-u

sam-s:


jorgenschaefer:
I do not understand this change. Why is this necessary or beneficial?

sam-s:
often I want to search for a string which is similar to an existing symbol but is not identical to it.
to do that now I need to edit the current buffer and hit `C-c C-s`.
sometimes I do not want to modify the buffer, so I do `C-u C-c C-s` and edit the current symbol in the minibuffer.

jorgenschaefer:
Ah, you add the symbol to the query, nice! Thank you!

Issue #1178, opened 2017-09-22
Start new shells from the library root


jorgenschaefer:
It would be less confusing if new shells were started with the current directory being the library root, just like tests are. See e.g. #1170 for an example of the confusion.

galaunay:
Do you think this should be done by default (seems fine for me, but this could confuse some users) or activated through an elpy option ?

jorgenschaefer:
I think default would be fine.

rgemulla:
Fixed with PR #1209.

Issue #1170, opened 2017-09-02
Elpy C-c C-c calls REPL from module directory


mashu:
Hi,

I have a problem with C-c C-c, it is not respecting project directory, but trying to execute my code from module directory with __init__.py file.

Here is full example https://github.com/mashu/ElpyExample and steps to reproduce.

Looking forward to solution,
Thanks a lot

jorgenschaefer:
Hello, and thanks for this really awesome and extensive report! I wish more people would file bug reports like this one :-)

Sadly, elpy right now just uses the current directory when starting a new shell. This is what some people want, but I understand why it's confusing. I created a new issue (see above) to change this behavior. For the time being, you can simply start the REPL from your project root by opening a file there and doing `M-x run-python` – future calls to the shell will use that one.

rgemulla:
Fixed with PR #1209.

PR #1209, opened 2017-11-04
Use project root as shell default directory

galaunay:
Fix issue #1178 (related to issue #1170).

New shells can now be started from the project root directory.
This feature is disabled by default to avoid confusing users and can be activated through the new option `elpy-shell-use-project-root`.

### DIFF
--- a/docs/ide.rst
+++ b/docs/ide.rst
@@ -196,6 +196,11 @@ The Shell Buffer
    Switch to buffer with a Python interpreter running, starting one if
    necessary.
 
+   By default, Elpy tries to find the root directory of the current project
+   (git, svn or hg repository, python package or projectile project) and
+   starts the python interpreter here. This behaviour can be suppressed
+   with the option ``elpy-shell-use-project-root``.
+
 .. option:: elpy-dedicated-shells
 
    By default, Elpy only starts a single interactive Python process. This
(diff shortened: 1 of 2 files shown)


galaunay:
Added the use of project root as default, following discussion in #1178.

jorgenschaefer:
Awesome, thank you. Could you just not add the option and make this the general behavior? I don't see why someone would not want this, and it just complicates the code more … Will merge as soon as you squash the commits :-)

galaunay:
Done and done.
I ran the tests, it does not seem to break anything.

rgemulla:
Is there still a way to disable using the project root? Personally, I do have multiple iPython scripts scattered around in a single project and I often do want the shell to start in the directory where the source file is located.

galaunay:
The first version I made (disappeared due to the commits squash) had an option to deactivate this behaviour.
I removed it to simplify the code and because we thought it was useless (see previous comments in this PR).
But If it breaks the workflow of some users, I think I need to re-add it.
At least if nobody has a better Idea to handle this...

galaunay:
Customizable option has been added back.

rgemulla:
Looks good to me.

Can you add a note to the appropriate section in the documentation (I just merged the updated interactive Python documentation for this reason; see ```docs/ide.rst```.)

galaunay:
I added a brief paragraph in the doc.
Nice work by the way.

PR #1208, opened 2017-11-03
elpy-shell: Non-blocking capturing of output, minor improvements, and bugfixes

rgemulla:
Fixes #1205, #1210, #1219. 

Also various smaller improvements in elpy-shell.

PR #1210, opened 2017-11-04
Fix dedicated-shells and echo-input conflict

galaunay:
Before that, activating both `elpy-dedicated-shells` and  `elpy-shell-echo-input` options leads to the creation of new python shells at each code evaluation.

Due to the use of `(elpy-shell-get-or-create-process)`, that return a new process when used while in the python shell buffer.

@rgemulla: Do you thing this would break something else ?

### DIFF
--- a/elpy-shell.el
+++ b/elpy-shell.el
@@ -502,7 +502,7 @@ Prepends a continuation promt if PREPEND-CONT-PROMPT is set."
   (let ((buffer (current-buffer)))
     (set-buffer (process-buffer (elpy-shell-get-or-create-process)))
     (let ((initial-point (point))
-          (mark-point (process-mark (elpy-shell-get-or-create-process))))
+          (mark-point (process-mark (get-buffer-process (current-buffer)))))
       (goto-char mark-point)
       (if prepend-cont-prompt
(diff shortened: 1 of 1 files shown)


rgemulla:
Good catch!

I think the culprit here is ```elpy-shell-get-or-create-process```. I suggest to leave the echoing code as it is, but to fix ```elpy-shell-get-or-create-process``` so that, when called from a Python shell buffer, returns the process associated with that shell. (Otherwise, bugs may occur at other places, too.)

Do you think this would be feasible?

rgemulla:
Also, for ```elpy-shell--append-to-shell-output```, it may be more suitable to start with
```lisp
(let* ((process (elpy-shell-get-or-create-process))
       (process-buffer (process-buffer process)))
  (with-current-buffer process-buffer
  ...
  )
```
and then use ```process``` and ```process-buffer``` throughout. This also avoids using the ```buffer``` variable.

rgemulla:
I fixed this in PR #1208.

galaunay:
I am closing this one then, 
thanks again for making interactive python easier on elpy.

galaunay:
I am closing this one then, 
thanks again for making interactive python easier on elpy.

PR #1211, opened 2017-11-04
Enhance shell reaction to code sending

galaunay:
### 1. Avoid the first output line to be directly printed after ">>>".
Evaluating:
```Python 
print('Hello world')
print('Happy to be here')
```
now gives:
```Python
>>>
Hello world
Happy to be here
>>> |
```
(`|` is the cursor)
instead of:
```Python
>>> Hello world
Happy to be here
>>> |
```
### 2. Gives a visual indication on the state of the shell (working or not).
While working (without outputs), the python shell will look like this:
```Python
>>>
|
```
instead of this:
```Python
>>> |
```
This gives an indication on the state of the python shell (working or waiting).

### DIFF
--- a/elpy-shell.el
+++ b/elpy-shell.el
@@ -789,7 +789,7 @@ Otherwise, skips forward to the next code line and sends the
 corresponding statement."
   (interactive)
   (elpy-shell--ensure-shell-running)
-  (elpy-shell--append-to-shell-output "\n")
+  (when (not elpy-shell-echo-input) (elpy-shell--append-to-shell-output "\n"))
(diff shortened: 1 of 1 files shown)


rgemulla:
Is this only done when inputs are not echoed? (If input is echoed, no additional newlines are needed.)

galaunay:
You are right, thank you, 
It is now only done when input is not echoed.

rgemulla:
Works flawlessly for me.

jorgenschaefer:
Thank you!

PR #1212, opened 2017-11-04
Avoid the delay of 3 seconds when sending code for the first time.

galaunay:
There is a hard-coded delay of 3 seconds when sending code for the first time (using `elpy-shell-send-region-or-buffer` for example).
It can be annoying if ones needs to restart the python shell often.
Moreover, 3 seconds may not be enough for slow computer or heavily loaded python startups.

This fix instead detects the display of the first prompt ('>>>') in the python shell buffer.
I couldn't find a better way of doing this (e.g. catch a signal from the python shell ?).

@rgemulla: did I understand this delay correctly ?

### DIFF
--- a/elpy-shell.el
+++ b/elpy-shell.el
@@ -309,11 +309,14 @@ Python process. This allows the process to start up."
 If the shell is not running, waits until the first prompt is visible and
 commands can be sent to the shell."
   (with-current-buffer (process-buffer (elpy-shell-get-or-create-process))
-    (let ((inhibit-text-field-motion t))
-      (while (progn
-               (goto-char (point-min))
-               (not (re-search-forward "^>>>" nil t)))
-        (sleep-for .1))))
+    (let ((cumtime 0))
+      (while (and (when (boundp 'python-shell--first-prompt-received)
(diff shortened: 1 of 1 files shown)


rgemulla:
On my system, the prompt is different. I think a more stable way is to use ```python-shell-first-prompt-hook```.  We can add a function to the hook, which tells ```elpy-shell-get-or-create-process``` to stop waiting. I'd nevertheless suggest to put a maximum waiting time in ```elpy-shell-get-or-create-process``` in case something goes wrong with prompt detection.

galaunay:
You are definitely right about this, thanks.
I assume you use Ipython, with the `In [1]:` prompt ?

Using`python-shell-first-prompt-hook` works, but I tried here a more simple implementation using the local variable `python-shell--first-prompt-received`.
I would be interested to know if it work for you @rgemulla.

rgemulla:
Yes, that's a simpler solution and much better than the timeout hack I had before. Works flawlessly for me. The warning can probably be dropped, since ```python.el``` warns itself when no prompt can be detected.

Issue #1214, opened 2017-11-12
"go to assignment" function?


cryptomaniac512:
I am not always need to go to definition, for this situations jedi implements "go to assignment". How can I use it with this package?

galaunay:
It's not currently implemented,
I made a pull request to add this feature #1216.
Any idea of a convenient keybinding for this function ?

cryptomaniac512:
Hm... Maybe `C-M-,`?

cryptomaniac512:
Hi there! What about this feature? =)

galaunay:
The code is written :), just waiting for @jorgenschaefer to review and merge it.
If you want it now, you can install elpy from the pull request branch:
[Add-goto-assignment-function](https://github.com/galaunay/elpy/tree/Add-goto-assignment-function).
If you do so and find any bug/unexpected behavior, please post them here #1216 so I can fix it.

cryptomaniac512:
Okay, thanks! You're cool! =)

jorgenschaefer:
The fellow did the same, sorry for the delay :-)

PR #1216, opened 2017-11-14
Add goto assignment function

galaunay:
Follow issue #1214.
Add `elpy-goto-assignment` function to benefit from the jedi `get_assignments` feature.

In this situation:
```Python
class TestClass(object):
    def __init__(a, b):
        self.a = a
        self.b = b

testclass = TestClass(2, 4)

testcla|ss.a
```
`elpy-goto-definition` jumps to the class definition:
```Python
|class TestClass(object):
    def __init__(a, b):
        self.a = a
        self.b = b

testclass = TestClass(2, 4)

testclass.a
```
while `elpy-goto-assignment` jumps to the variable assignment:
```Python
class TestClass(object):
    def __init__(a, b):
        self.a = a
        self.b = b

|testclass = TestClass(2, 4)

testclass.a
```

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -1591,6 +1591,14 @@ with a prefix argument)."
         (elpy-goto-location (car location) (cadr location))
       (error "No definition found"))))
 
+(defun elpy-goto-assignment ()
+  "Go to the assignment of the symbol at point, if found."
+  (interactive)
+  (let ((location (elpy-rpc-get-assignment)))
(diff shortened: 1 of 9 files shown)


jorgenschaefer:
This is awesome, thank you so much. Especially for also writing the tests.

galaunay:
I would have felt bad to make the coverage score drop :).

Issue #1225, opened 2017-11-28
Command C-u C-c C-c does not work


DungxNguyen:
Hi, I can not make C-u C-c C-c to work. I tried to debug the problem but I can not find the function
`elpy-shell-send-region-or-buffer` in elpy-shell.el, though I know it is bound to C-c C-c.
Otherwise, I find the function 
`elpy-shell--send-region-or-buffer-internal`
is actually called each time I press C-u C-c C-c from the function
`elpy-shell-send-region-or-buffer-and-step`

And I can see that the function `elpy-shell-send-region-or-buffer-and-step` does not accept any prefix argument, so the argument can not be passed to the shell, which means the argument `arg` is always nil even I press C-u prior to the C-c C-c.

My emacs version is 25.3. Elpy version 20171124.720 from melpa.

DungxNguyen:
Can somebody tell me where the function `elpy-shell-send-region-or-buffer` is?

rgemulla:
The function ```elpy-shell-send-region-or-buffer``` is auto-generated by line

```elisp
(elpy-shell--defun-step-go elpy-shell-send-region-or-buffer-and-step)
```

in ```elpy-shell.el```. The prefix argument is currently not passed along; I'll push a fix as soon as I can find some time. (It would be helpful if you could provide a short example to test it.)

Until the fix is in, you can use ```elpy-shell--send-region-or-buffer-internal```, which does accept the prefix argument, but does not support the input/output echoing features of elpy-shell (which is what is added by ```elpy-shell-send-region-or-buffer```).

rgemulla:
I created a PR that fixes this issue (#1226).

rgemulla:
Pushed the PR. This should fix this issue.

PR #1226, opened 2017-11-29
elpy-shell: bug fix and improvements

rgemulla:
Fixes bug #1225.

Also adds functionality to evaluate code cells. This can be used to send custom sets of lines to the shell (by default, the set of lines is delimited by ``###`` lines). Full documentation is in PR #1227.


elpy-shell.el
```diff
@@ -884,71 +915,118 @@ below point and send the group around this statement."
...
+        (end (save-excursion
+               (forward-line)
+               (if (re-search-forward elpy-shell-cell-boundary-regexp nil t)
+                   (forward-line -1)
+                 (end-of-buffer))
+               (end-of-line)
+               (point))))
+    (if beg
+        (progn
+          (elpy-shell--flash-and-message-region beg end)
```
galaunay: Is it not possible to use `elpy-shell-send-region-or-buffer` here to avoid redundancy ?

galaunay:
Is this a wanted behaviour to send the code until the end of the buffer in there is not ending block ?

Minor issue maybe, the code block is not sent if the cursor is on the code block ending line.
i.e. on the third line here:
```Python
###
print("Hello world")
###
```

elpy-shell.el
rgemulla: Possible, but then we'd have to store, change, and restore the region. I'd vote to keep it as is. The entire code-sending parts should probabily be simplified into:

1. Navigation commands for start/end
2. Code sending commands (automatically build from navigation commands)

I'll do this at some point. The main reasoning is that users can then also use the navigation commands directly (e.g., to jump to the next codecell etc.)

rgemulla:
Yes. The semantics is: The current cell starts at the cell boundary on or above the line of point and ends at the cell boundary below point (or end of buffer, if none). A cell is executed only if the starting boundary matches a codecell beginning. In your example above, the second ```###``` acts as a cell boundary and as a beginning of a new codecell.

This makes it easy, for example, to support the cells of a notebook exported from Jupyter directly (where the beginning is "# In[.*]:" or "# <codecell>"). It also avoids accidential execution of non-codecells (where the begining is "# Out[]:" or "# <markdowncell>" etc.). This is the default behaviour.

What is boundary and what codecell beginning can be customized via the regexps.

galaunay:
I understand now, thank you for the clarifications.

rgemulla:
To me, the PR is ready. @galaunay Agreed?

elpy-shell.el
galaunay: This indeed seems a good idea is the number of functions sending code parts increases.

galaunay:
The mix between bug fix and code cell sending make it a little hard to read, but this seems ok for me.

rgemulla:
Yes, I'll try not to do that again. Thanks!

PR #1236, opened 2017-12-08
avoid "Unused lexical argument" warnings

sam-s:


rgemulla:
I find it somewhat unsatisfying to change outside visible variable names to get rid of byte compile warnings. But I have no strong option either way.

gopar:
Seems this is somewhat common: https://github.com/bbatsov/emacs-lisp-style-guide#naming

galaunay:
I though several times about trying to reduce compilation warning number, 
thank you for having the courage to actually do it.

I think this is a valuable information to know which parameter as a real influence on the function,
I am voting in favor of this.

gopar:
Has my vote. It can be merged.

PR #1242, opened 2017-12-10
Remove importmagic

### LINKED ISSUE #1230: Remove importmagic
importmagic creates a lot of problems because it tries to import all modules. Many modules do not know how to deal with this, and either emit spurious output, cause CPU hogging or even crash Python altogether.

As sad as it is, since importmagic adds some nice features, in the current state of affairs it's not a good solution. At the bare minimum, the suggestion for installation should be removed. Probably all code related to importmagic can be removed as well.

gopar:
Closes #1230


gopar:
Yey, coverage increased \o/

galaunay:
90% is a step, congratulation :).

I greped an occurrence of 'importmagic' in .travis.yml.
I suppose this can be safely removed as well.

gopar:
Thanks. Weird, I was using projectile to grep through the project but that one never showed up

gopar:
Something like `(make-obsolete 'importmagic-functions "support for importmagic has been dropped")`?

galaunay:
Yes, I suppose.
At least for the interactive functions ?

jorgenschaefer:
If you add compatibility functions, you can leave the keybindings there for a release or two. Otherwise, I would not bother with compatibility functions. It's just stuff to clean up later.

gopar:
I added the keybindings back. So, now whether the user uses the keybindings or calls the function interactively, they should see the obsolete message

galaunay:
Seems fine to me

gopar:
Alright, if I get an additional thumbs up then I'll merge

jorgenschaefer:
:+1:

eb-enzo:
Thanks for your quick response. I'll take a look at `importmagic.el`. However, I mainly used `importsmagic` for imports ordering and formatting, not for resolving unknown symbols or adding new ones. Maybe I should search for a lightweight solution for my problem.
Thanks again.

jorgenschaefer:
I can highly recommend [isort](https://pypi.python.org/pypi/isort) for that :-)

eb-enzo:
EXACTLY what I need. Thanks so much.

PR #1243, opened 2017-12-11
Add python3.6 to travis ci

gopar:
Does anyone know if there is a strong usage of 3.3? From what I've heard most people are on 3.4 or higher if they're doing py3 stuff

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ python:
   - 3.3
   - 3.4
   - 3.5
+  - 3.6
 env:
   # - EVM_EMACS=emacs-24.1-travis
   # - EVM_EMACS=emacs-24.2-travis


gopar:
Actually thought this was gana pass without testing 😆 

Looks like this needs work, which I can do. Should elpy support 3.6? I think yesssssss

jorgenschaefer:
Yes please! :-)

galaunay:
Apparently they are not real errors but tests that need to be adapted to python3.6.

I launched Elpy tests with python 3.6.3 and I got additional errors (notably one linked to rope refactoring that does not work anymore).
Should we add support for python 3.6.* ?

jorgenschaefer:
> Should we add support for python 3.6.* ?

Yes please! :-D

If Rope is making problems, we really can even get rid of it entirely or disable it for whole versions. There was an attempt to revive its development, but I think that died off, too.

galaunay:
I also noticed that emacs 25.3 was missing from the travis file.
@gopar is it possible to add python3.6.1, python3.6.2, python3.6.3 and emacs25.3 to this PR ?
@jorgenschaefer I will try to fix the issues and fall back to removing/disabling rope.
Is there any real advantages to use rope instead of jedi btw ?

jorgenschaefer:
> Is there any real advantages to use rope instead of jedi btw ?

Tricky. I don't think anymore, no. There used to be some things rope could find that Jedi did not, but I don't think that's a real thing anymore. The only thing Rope is really good for still is the refactoring stuff.

jorgenschaefer:
I think it would be ok to keep rope just for the refactoring functions, but get rid of rope for the IDE stuff (completion, find symbol, etc.), right?

gopar:
Yeah, I still use rope for refactoring small things but that's about it. I don't mind it being removed  :P 
But I also agree with @jorgenschaefer that removing the IDE stuff would probably be a better start to getting users to depend on jedi primarily

galaunay:
Argh,
my bad, I just realized I was running the tests with the wrong virtual environment...
There is nothing wrong with rope refactoring.
Putting python3.6 in .travis.yml should be enough, no need for python3.6.1, python3.6.2 and python3.6.3.

@gopar I made this [commit](https://github.com/galaunay/elpy/commit/4e87ad23bf02c535f2acef220659dc2e29d7832f) to fix the errors we got with python3.6, if it can be helpful.

 I  still made a PR (#1245) for removing rope IDE features, we can discuss it there.

gopar:
@galaunay sweet thanks! But I think we should wait until #1245 is merged. No point in fixing tests for a non-existing feature

galaunay:
@gopar #1245 was merged.
Let's see if it still fail after a rebase :).

jorgenschaefer:
@gopar @galaunay I do not know where to put this, so this will have to do, but I really can not express how much you are doing here for me with all the awesome work you do. This really helps me, and of course Elpy, so much! Thank you.

gopar:
@jorgenschaefer it's mostly @galaunay doing all the work \o/

I was wondering if there is like a chat that collaborators can use to discuss things that are not necessary related to issues? For example I want to know what would be the criteria for being able to close issues and such? Just questions like that, that don't really belong in an issue.

galaunay:
Really happy to help.
Hope Elpy will remain the great package it is for a long time !

@gopar This is a really good idea, not the first time I could have used such thing.
But I don't think github offers anything that can help... 
mail ?

gopar:
Looks like some github projects use https://gitter.im/ as a general chat room. Might be worth looking into?

galaunay:
Yes, great idea !
I think Gitter has some kind of github integration, 
do you know exactly how it work ?

gopar:
Just tried it, it's pretty simple to create a room. It even gave me the option to create one for this repo as well. @jorgenschaefer are you okay with using that?

Issue #1107, opened 2017-03-24
Remove explicit IPython support


jorgenschaefer:
IPython 5 has changed a lot, and made integration with Emacs rather difficult. There have been a lot of error reports because of this. Elpy does not add much to basic Emacs support of IPython, and with the IPython 5 changes, using IPython in a comint buffer is getting a worse and worse idea.

Elpy should just do away with the explicit IPython support, and pass IPython error reports on to Emacs core.

ltsampros:
Maybe open this issue in ipython's github issue tracker as well?

Their documentation states explicit support and offers advice as well.

rgemulla:
What do you mean by dropping IPython support, i.e., what kind of support? I am using it with Emacs+elpy without any issues. (And to me, using it in a comint buffer with Emacs is a great idea ;))

jorgenschaefer:
Elpy offers `elpy-use-ipython` which sets the environment to use ipython as the interactive shell. This has caused a lot of support requests, because ipython changed so much and seemed not designed to work in a comint buffer anymore (requires some semi-documented environment variable to work, for example). If those support requests are not an issue for you, I don't have any particular stakes in this. :-D

rgemulla:
Hmm, most of the code can go now since we require Emacs 24.4 (at some point, we should probably start removing such dead code).

I am on windows, but I am not using ```ipython-script.py```; not sure if this is still recent.

I am still in favor of keeping it, but simplifying to essentially setting the env variable, the interpreter and its argument. We can also have similar functionality for Jupyter console. If you agree to this, I can make a PR for this.

jorgenschaefer:
I'm happy to keep the direct support if we have someone who can work on it. I don't use ipython at all :-D

galaunay:
I don't use it either. 
I cannot be of any help here.

gopar:
I don't use it as well

rgemulla:
As there is not much to setup anymore, I suggest to drop the function (which I don't use myself as well) and write a "how to setup" entry in the "interactive python" section of documentation (where we should also recommend jupyter console, see #1152). I'll take a shot at this when I find some time.

rgemulla:
See #1244. I also removed ``elpy-use-cpython`` because it doesn't do anything anymore (all code was for prior Emacs versions).

PR #1244, opened 2017-12-11
elpy-shell: removed setup functions, added documentation

rgemulla:
The ``elpy-use-ipython`` and ``elpy-use-cpython`` functions don't do anything particularly useful anymore now that Emacs 24.4 is required. The PR removes these functions but adds documentation for how to set up common environments. See also #1107.


rgemulla:
Some test fail now because ``elpy-use-cpython`` is gone. My suggestion would be to simply remove this function anywhere in the tests (it doesn't do anything anymore). Before going ahead, I'd like to hear your thoughts on this though.

rgemulla:
I had a quick look, it would actually suffice to remove ``elpy-use-cpython-test.el``.

galaunay:
This seems fine to me.
Just by curiosity, have you any ideas why there is problem with ipython, but not with ipython in jupyter ?

rgemulla:
No, I am not sure why.

rgemulla:
@jorgenschaefer Are you also fine with removing both setup functions?

jorgenschaefer:
Hm. I'm all in favor of doing this, though I think it would be nice to deprecate those functions with a reference to the documentation for a few versions, so people don't get too confused?

rgemulla:
Yes, that's a good idea. One option would be to add the functions again; the only thing they do then is to throw an error with a reference to the documentation. Is this what you mean?

An alternative is to mark them obsolete, but I am not sure if elpy users would even realize.

jorgenschaefer:
> One option would be to add the functions again; the only thing they do then is to throw an error with a reference to the documentation. Is this what you mean?

Yes, that's what I mean. Sorry for the confusion :-)

PR #1245, opened 2017-12-12
Remove Rope IDE features

galaunay:
Following discussion in #1243.
Remove Rope IDE features and make Jedi backend the default one.


gopar:
Sweeeet, Can you resolve the conflict? Wanna see the tests run :P

galaunay:
Some rebase later... :)
I think it's good, but it's a bit tricky because half the rope-related stuff need to be preserved for refactoring functions.

jorgenschaefer:
lgtm, thank you very much! The whole backend code and the tests can likely be simplified a lot after this. And I guess the "native" backend can go away as well and elpy will just not provide completion without jedi, simplifying the code even further. (Not necessarily in this PR, but it's something that can be done in the future :-) )

galaunay:
Great ideas, I will try to clean all of this a little in the following days.

galaunay:
Made my best.
I assume the code might still be simplified in 'server.py', but I am unsure how.
If someone has some insight on this...

jorgenschaefer:
We can also simplify more later. Do not hesitate to merge if it works. :-)

galaunay:
Noted.
600 lines less is still a good start :)

PR #1248, opened 2017-12-17
Append wildcard to directory patterns from `completion-ignored-extensions`.

vytas7:
If GNU find ``-iwholename`` patterns end with a trailing slash, they are effectively ignored, and a warning is emitted.


jorgenschaefer:
Thank you for this fix! I did some minor code style changes and applied to master.

PR #1253, opened 2017-12-27
Support case-insensitive completion for company backend

YorkZ:
Hi Jorgen,

Thank you very much for this wonderful package.

I noticed that completion was case sensitive, so I added the support to allow users to configure company backend for case-insensitive completion.

York

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -2779,6 +2779,12 @@ If you need your modeline, you can set the variable `elpy-remove-modeline-lighte
 ;;;;;;;;;;;;;;;;;;;
 ;;; Module: Company
 
+(defcustom elpy-company-ignore-case nil
+  "Non-nil if elpy-company-backend returns case-insensitive
+matches."
+  :type 'boolean
+  :group 'elpy)
+
 (defun elpy-module-company (command &rest _args)
   "Module to support company-mode completions."
   (pcase command
@@ -3048,6 +3054,7 @@ and return the list."
     ;; duplicates => t if there could be duplicates
     (`duplicates
      nil)
+    (`ignore-case elpy-company-ignore-case)
     ;; no-cache <prefix> => t if company shouldn't cache results
     ;; meta <candidate> => short docstring for minibuffer
     (`meta
--- a/elpy.el
+++ b/elpy.el
@@ -3008,8 +3008,8 @@ and return the list."
              for pytel-cand = (replace-regexp-in-string "($" "" pytel-cand)
              for pytel-cand = (replace-regexp-in-string "^.*\\." ""
                                                         pytel-cand)
-             if (not (member pytel-cand candidate-names))
-             do (push (list (cons 'name pytel-cand)) candidates)))
+             unless (member pytel-cand candidate-names)
+             collect (list (cons 'name pytel-cand)) into candidates))
           candidates)))))
 
 (defun elpy-company-backend (command &optional arg &rest ignored)


jorgenschaefer:
Thank you for this contribution! What exactly does "completion is case sensitive" mean, i.e. what misbehavior are you seeing? I think this never had a problem as I configure Emacs to be case insensitive in general.

Does this setting cause any problems when completing mixed-case symbols?
E.g. does it complete the wrong case when I type the wrong case?
Or will company then not allow the completion of either `FOO` or `foo` if both are available?

If neither, is there any reason why someone would not want this to happen? I.e. could the option be removed and this always be set to `t`?
If either of those is the case, is there a better solution?
In any case, if the option variable remains, it needs a better docstring.

YorkZ:
Sorry I should have explained it in more detail. Let's take a look at the following code:

    class FooClass:
        pass


When completion is case sensitive, if you type the following:

    my_var = foo

No completion will occur. The completion candidate `FooClass` will show up only if you typed:

    my_var = Foo

However, if completion is case insensitive which is all  this commit is about, when typing the same text, the completion candidate `FooClass` will pop up.

Regarding the docstring, sorry if it's not clear but I was just borrowing the "terminologies" from the docstring of `company-backends` variable in company.el:

    `ignore-case': Return t here if the backend returns case-insensitive
    matches. This value is used to determine the longest common prefix (as used
    in `company-complete-common'), and to filter completions when fetching them
    from cache.

I agree it might not be very clear, can you suggest a better description?

Making completion case intensive will result in more completion candidates, but the user doesn't have to type Capital letters like in this example. Somebody like case intensive completion; others, case sensitive completion. Therefore, I think it's better to make it customize-able.

Hope my explanation helps clear thing up a bit.

jorgenschaefer:
Does this mean you can still complete `foo` and `FOO` separately if both exist?

gopar:
Hmm, that's weird. I get case insensitive completions already:

screen shot 2017-12-28 at 11 11 40 am

YorkZ:
@jorgenschaefer Let's say the above code includes another class, i.e.:

    class FooClass:
        pass

    class foo_class:
        pass

If completion is case sensitive, when you type `foo`, you get `foo_class` but not `FooClass` as candidates. If you want to get `FooClass`, you have to type `Foo`. On the other hand, with case insensitive completion, when you type `foo`, you get both `foo_class` and `FooClass` as candidates, and then you decide which one you want. In other words, case sensitive completion is more "strict" and will results in less candidates which is good, but you have to type capital letters which is not so good. Whereas case insensitive completion is less "strict", you will get more candidates (not good) but you don't have to type capital letters (good). Like I said, it's a matter of personal preference, and this commit just gives people a choice so that different people can get different behavior from the package. I believe if you  pull this change to your system and tried it out yourself, you will immediately know what it's all about. Sometimes it's really hard to describe things in language.

YorkZ:
@gopar Strange, I guess you might have other configuration. Can you test it with `emacs -Q`?

YorkZ:
@jorgenschaefer 

> Does this mean you can still complete foo and FOO separately if both exist?

Yes.

jorgenschaefer:
If you can complete both `foo` and `FOO` if both exist, I'd prefer doing this as the standard without any custom variable needed. Fewer options is better in my book.

Why gopar and I do not see the case-sensitive behavior would be interesting still, though.

sawan:
> Why gopar and I do not see the case-sensitive behavior would be interesting still, though.

Definitely some setting on your side; I get the case-sensitive behavior.

PR #1256, opened 2018-01-01
fix #1255, `elpy-xref--identifier-at-point` should return a string

### LINKED ISSUE #1255: xref-backend-identifier-at-point returns a symbol but xref expects a string
This causes issues, e.g. one way to reproduce:

1. Open python source
2. Move point to an identifier
3. Run xref-find-references
4. Select default (M-n)

Error is thrown: `let: Wrong type argument: char-or-string-p, num_val`

Solution: Have `elpy-xref--identifier-at-point` return a string.

E.g.

```elisp
(defun elpy-xref--identifier-at-point ()
    "Return identifier at point."
    (symbol-name (symbol-at-point)))
```

galaunay:
This fix the issue #1255 reported by @rksm.

`elpy-xref--identifier-at-point` now return a string instead of a symbol, in agreement with the xref documentation.

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -3115,7 +3115,7 @@ Points to file FILE, at position POS."
     "Return identifier at point.
 
 Is a string, formatted as \"LINE_NUMBER: VARIABLE_NAME\"."
-    (let ((symb (substring-no-properties (symbol-name (symbol-at-point))))
+    (let ((symb (thing-at-point 'symbol t))
           (assign (elpy-rpc-get-assignment)))
       (when assign (format "%s: %s"
                         (line-number-at-pos (+ 1 (car (cdr assign))))


elpy.el
```diff
@@ -3113,7 +3113,8 @@ Points to file FILE, at position POS."
 
   (defun elpy-xref--identifier-at-point ()
     "Return identifier at point."
-    (symbol-at-point))
+    (when (symbol-at-point)
+        (substring-no-properties (symbol-name (symbol-at-point)))))
```
jorgenschaefer: Considering `symbol-at-point` basically does `(thing-at-point 'symbol)`, which in turn returns a string, I would suggest using that directly.

galaunay:
You are right...
Closing this and continuing discussion in #1255 to find the right reason of this issue.

galaunay:
Just to be sure,
does `(symbol-at-point)` not returning a symbol, due to the use of `intern` ?
This is the implementation of `symbol-at-point` I got with emacs 25.3:
```Lisp
(defun symbol-at-point ()
  "Return the symbol at point, or nil if none is found."
  (let ((thing (thing-at-point 'symbol)))
    (if thing (intern thing))))
```

galaunay:
made obsolete by #1277.

PR #1258, opened 2018-01-01
Add more complete identifiers for xref

### LINKED ISSUE #1257: Xref identifiers need to carry information on position
## Problem:
`xref-find-references` does not work as expected when variable names get redefined.
For example, in the following situation (| being the cursor)
```Python
var1 = (1, 2)
var|1 = 2
var2 = var1 + 2
```
`xref-find-references` will only find the first line, because jedi make a distinction between `var1` on the first line and `var1` on the second and third line (after its redefinition).

## Solution:
Identifiers used by elpy to communicate with xref (`elpy-xref--identifier-at-point`) currently only carry information about the variable name.
Adding information on the variable position in the buffer would allow to differentiate two variables with the same name, and fix this issue.

I will try to make a PR in the following days addressing this.

galaunay:
Fix #1257.

## Problem:

`xref-find-references` does not work as expected when variable names get redefined.
For example, in the following situation (`|` being the cursor):
```Python
var1 = (1, 2)
var|1 = 2
var2 = var1 + 2
```
`xref-find-references` will only show the first line, because jedi makes a distinction between `var1` on the first line and `var1` on the second and third line (after its redefinition).

## Solution:

Identifiers used by elpy to communicate with xref (returned by `elpy-xref--identifier-at-point`) currently only carry information about the variable name.
This PR adds information on the line number of the variable assignment, and allows to make a distinction between different variables sharing the same name.

## Testcases:

In this situation:
```Python
var1 = (1, 2)
var|1 = 2
var2 = var1 + 2
```
`xref-find-references` will now show `var1` on the second and third line,
and in this situation:
```Python
va|r1 = (1, 2)
var1 = 2
var2 = var1 + 2
```
It will show `var1` on the first line only.


jorgenschaefer:
Interesting idea, thank you for the implementation!

Does this work if the identifier is used on the same line? Say

```Python
var1 = (1, 2)
va|r1 = 2
var1 = var1 + 2
```

galaunay:
Yep, in this example:
```Python
var1 = (1, 2)
va|r1 = 2
var1 = var1 + 2
```
`xref-find-reference` now makes the distinction between the first occurrence of `var1` (line 1), the second one (line 2 and line 3, right side of the equal sign), and the third one (line 3, left side of the equal sign).
Consequently, at the cursor position, `xref-find-reference` gives:
```
/tmp/test.py
  2:	var1 = 2
  3:	var1 = var1 + 2
```

jorgenschaefer:
lgtm then, please merge as you see fit :-)

galaunay:
I made a few tests with bigger files to check that everything was fine.
and its merged !

jorgenschaefer:
Thank you for your work! :-)

PR #1259, opened 2018-01-05
Clean up jedi backend

gopar:
Remove special cases of jedi failing, since most of those issues have
been fixed in the newer releases

### DIFF
--- a/elpy/jedibackend.py
+++ b/elpy/jedibackend.py
@@ -143,17 +143,6 @@ class JediBackend(object):
             call = None
         if not call:
             return None
-        try:
-            call.index
-        except AttributeError as e:
-            if "get_definition" in str(e):
-                # Bug #627 / jedi#573
-                return None
-            elif "get_subscope_by_name" in str(e):
-                # Bug #677 / jedi#628
-                return None
-            else:
-                raise
         return {"name": call.name,
                 "index": call.index,
                 "params": [param.description for param in call.params]}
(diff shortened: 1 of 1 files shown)


gopar:
So the failed tests didn't really failed. They just never started. Is it possible to just run those specific ones?

gopar:
Nvm, fixed it. And we're back up in the low 90's \o/ 😄

jorgenschaefer:
Excellent, thank you!

Do you think it's a good idea to remove the tests? Keeping them would protect us from regressions. Though I think it would be good if Jedi had these tests and not us, that's not how the world is. (Serious question, I can see arguments both ways.)

gopar:
I wouldn't mind having the tests but when I made the first commit (of just removing the jedi code and leaving the tests) I got this result in [CI](https://travis-ci.org/jorgenschaefer/elpy/builds/325286787?utm_source=github_status&utm_medium=notification). I didn't really dig into but almost all of the py2.7 tests fail with 2 specific tests being the issue. Only of instance of py2.7 didn't fail.

Also, most of the issues are fixed via the latests jedi version? At least when I was looking up the issues in the jedi repo, all of them were closed

jorgenschaefer:
Well, it seems then that the "bad escape sequence" error was not fixed in Jedi and still needs to be ignored by Elpy?

I do not know which of these errors were actually fixed. Jedi tends to close a bunch of reports when a new release comes out because "rewrote most of this, should be fixed" :-D

gopar:
Sounds like you want to keep the tests then? :D

jorgenschaefer:
Yes … I think removing the workarounds from the actual code is great when they're not necessary anymore, but keeping the tests is I think a good strategy with a volatile dependency like Jedi.

gopar:
So I left the tests there to catch regressions, and left that one specific bug case in the code since it apparently isn't fixed in jedi yet. If it looks good, then I'll merge \o/

jorgenschaefer:
lgtm. Awesome, thank you! :-)

PR #1265, opened 2018-01-11
fix void function call in elpy-django-command

BackwardSpy:
When running `C-c C-x c (elpy-django-command)` I see the following:
`elpy-django--get-commands: Symbol’s function definition is void: -remove`
This change proposal appears to fix the error.

As with my other PR - I'm not very experienced with elisp so it's possible I'm missing something.

### DIFF
--- a/Cask
+++ b/Cask
@@ -11,6 +11,5 @@
 (depends-on "s" "1.11.0")
 
 (development
- (depends-on "dash")
  (depends-on "f")
  (depends-on "ert-runner"))
--- a/elpy-django.el
+++ b/elpy-django.el
@@ -123,7 +123,7 @@ require arguments in order for it to work."
     ;; get a list of commands from the output of manage.py -h
     ;; What would be the pattern to optimize this ?
     (setq dj-commands-str (split-string dj-commands-str "\n"))
-    (setq dj-commands-str (-remove (lambda (x) (string= x "")) dj-commands-str))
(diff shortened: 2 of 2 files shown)


galaunay:
Thank you for reporting this.
You are right, there is a problem here.

But I don't think your fix will work, 
`-remove` is a function from the dash library that does not do exactly the same thing as `remove` (you can check that by using `describe-function`).
Using `remove-if` in place should work though.

BackwardSpy:
Ah, I see what you mean. I did not know about dash. I have updated my branch to use `cl-remove-if` instead, and as far as I can tell it is doing what it's meant to do.

I noticed that the project's Cask file depends on dash. Is dash an older requirement that has since been dropped, or something else?

galaunay:
As far as I know, elpy requires dash only for development (some tests should have use dash at some points, I guess).
But after removing it, tests still run fine.
to avoid multiplying PR, would you mind removing the dash line in the cask file here ?

BackwardSpy:
Dash is now gone from the Cask file. :+1:

gopar:
Thanks a lot! Can't believe no one has caught this yet. I guess not a lot of ppl are using or if they are, they already have dash installed (Like me when I was writing it). I'll checkout the branch and test it out but generally looks good!

@galaunay I did a quick grep (dash functions seem to start with `-` such as `-remove`) and didn't find any other functions that matched that. Again I only did a simple grep

gopar:
Tested it and works. Merging \o/

PR #1277, opened 2018-01-31
Fix Xref reference mis-behavior

galaunay:
Some bugs were introduced by the last modification of the xref backend (#1258, my bad).
Notable `xref-goto-definition` and `xref-find-reference` now fail when the symbol assignment is in another file.

 This PR fix those bugs.
(also fix the documentation for `xref-find-references`).


jorgenschaefer:
Awesome, thanks!

Issue #1275, opened 2018-01-26
Use the RPC interpreter to run syntax checks


jorgenschaefer:
See #1162 – we install yapf etc. for the RPC interpreter, not the shell interpreter.

galaunay:
I don't get the issue.
Yapf is called by `elpy-rpc`, so should use `elpy-rpc-python-command`,
and installation is coherent, as it also uses `elpy-rpc-python-command`.
Am I missing something ?

jorgenschaefer:
`elpy-format-code` looks for the `yapf` executable, not the `yapf` module, for example. For syntax checks, we call `flake8` as a command, not the module. Etc. But we check the presence as modules. It's pretty inconsistent, and that is pretty confusing. Especially when you have a python interpreter set, but not the PATH for the installed scripts.

galaunay:
Got it.
Thanks

galaunay:
Was fixed by #1281.
I am closing this.

PR #1281, opened 2018-02-03
Check for yapf and autopep8 modules instead of executables

galaunay:
Following #1275.
Presence of yapf and autopep8 is now checked by verifying the presence of the modules in the RPC instead of verifying the presence of their binaries.

Additionally, is there a good reason for the syntax checker to be specified as an executable instead of as a module ?
Would it not be more coherent to use the module, as for yapf, autopep8, ...

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -949,6 +949,11 @@ item in another window.\n\n")
 
     ))
 
+(defun elpy-config--package-available-p (package)
+  "Check if PACKAGE is installed in the rpc."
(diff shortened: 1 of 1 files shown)


jorgenschaefer:
Looks good to me.

> Additionally, is there a good reason for the syntax checker to be specified as an executable instead of as a module ?

Not sure if this is what you mean, but yes, I think `elpy-syntax-check-command` should be `python -m flake8` :-)

PR #1286, opened 2018-02-05
Add missing dependency, fix #1285

### LINKED ISSUE #1285: Error msg on installation
Hello, 
I just installed elpy by following the steps given in readme.
I came upon an error in the *Compilation* log after launching 'package-install' 'elpy' command in emacs.
The error was 
elpy-shell.el:977:1:Error: Symbol's function definition is void: string-remove-suffix

I joined the full log since there were many warnings.
My config : Ubuntu 16.04, GNU emacs 24.5.1, python 3.6.1, anaconda 4.4.0, pip 9.0.1, with jedi, flake8, autopep8 and yapf installed just before.

I'm quite new with emacs, and thus I have no idea about the consequences of it ; i just know that an error is generally quite a bad sign ....
Thanks for the help !

[log_compil_err.txt](https://github.com/jorgenschaefer/elpy/files/1695647/log_compil_err.txt)

galaunay:
Fix Issue #1285.

## Problem
When installing elpy from melpa on a clean emacs, 
compilation raise the following error: 
```
elpy-shell.el:977:1:Error: Symbol's function definition is void: string-remove-suffix
```

## Solution
It is fixed by adding the dependency to `subr-x` in `elpy-shell.el`.

Still, I thought that requiring `subr-x` in `elpy.el` would be enough, 
any ideas why it is not the case ?

### DIFF
--- a/elpy-shell.el
+++ b/elpy-shell.el
@@ -24,6 +24,8 @@
 ;;
 ;;; Code:
 
+(eval-when-compile (require 'subr-x))
+
 ;;;;;;;;;;;;;;;;;;;;;;
 ;;; User customization
 


raynalm:
Thanks a lot, I'm glad the bug was reproducible.
I'll proceed with a new elpy install, should be fine now.

Apteryks:
Hello, I'm not sure if it matters much, but shouldn't the symbol made available using: ```(eval-when-compile (require 'subr-x))```

According to http://emacsredux.com/blog/2014/02/02/a-peek-at-emacs-24-dot-4-new-string-manipulation-functions/, "all functions in the library are defined as inline, so you don't need to require the library at runtime".

galaunay:
@Apteryks thanks for the article.
I made the modification.

jorgenschaefer:
You people are so awesome, thank you for all your work :-)

Issue #1276, opened 2018-01-29
[Question] - parallel documentation a la 'Scientific Mode'


Nicholas-Mitchell:
This is first of all a question, then perhaps later a feature request.

I can look up any documentation (docstrings) that Elpy can find using the simple `elpy-doc`  with the cursor over the object of interest.

I have recently tried out the PyCharm take on things and have discovered the Scientific Mode, whereby the corresponding documentation for wherever the cursor is, is automatically shown in a separate 'buffer'. Here is the examples on [their relevant release video](https://youtu.be/OHwh0c8UsW4?t=1m15s). While I can hit `C-c C-d` as often as I like, I really like the way PyCharm gives the documentation according to the current state of auto-complete.

For example, if I am at `numpy.random.` - the documentation is shown for wherever my cursor then resides on the list of possible completions. this allows me to much more quickly decide which completion I need - it does not interrupt my flow of thinking.
(The other features shown in that Data Science section of the video would be nice to have, like plotting outputs directly in the same buffer etc, but Emacs itself has some limitations there, I know).

Given how quickly the `elpy-doc` call works, I can't imagine it would be impractical to have it running, say every 0.5 seconds.

I get the feeling this might be simple to implement, but I am sadly not as proficient in E-lisp as required to build this myself.

Does anybody have thoughts on this? Would anybody else want this?

galaunay:
For displaying documentation based on what is under the cursor,  
I guess the way `company-idle-delay` works is a good example of how to do it (notably the `company-post-command` function, that use `run-with-timer`, if you want to take a look and learn a bit of elisp :) ).

Though, on some of my (old) computers, `elpy-doc` can sometimes be quite slow (several seconds) to extract the documentation from my packages. 
So running elpy-doc as an asynchronous process (like pycharm does it) seems necessary to avoid having Emacs freeze each second to get the documentation.
It does make implementing this more difficult.

Anyway, it is an interesting feature, I will give it a try (unless you want to), see how it goes.

Regarding having documentation during completion, you should take a look at [company-quickhelp](https://github.com/expez/company-quickhelp), it does exactly that, but in a popup instead of in an adjacent window.

Nicholas-Mitchell:
@galaunay - thanks for implementing it!

Could you please tell me the best way to try it out? Just copy the relevant code from `elpy.el` or will it be necessary to clone the whole repo? If so, how do I do that without touching my current setup at all? Thanks in advance  :smile:

[...]

PR #1288, opened 2018-02-07
Auto-update of the documentation buffer

galaunay:
Following feature request #1276.

## Summary
This PR adds the automatic update of the documentation buffer based on the symbol at point.
Documentation is also updated in company, based on the currently selected candidate.

To avoid documentation poping up unexpectedly, the documentation is only updated when the documentation buffer is visible (i.e. associated to a window).

## Control of the feature
The delay before which updating the documentation is controlled by the option `elpy-autodoc-delay`, that can be set to `nil` to deactivate the feature.

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -2138,7 +2138,7 @@ If nil, deactivate the documentation automatic refresh."
     (`post-command
      (when elpy-autodoc-delay
        (when elpy-autodoc--timer
-         (cancel-timer elpy-autodoc-timer))
+         (cancel-timer elpy-autodoc--timer))
        (setq elpy-autodoc--timer
              (run-with-timer elpy-autodoc-delay
                              nil
--- a/elpy.el
+++ b/elpy.el
@@ -75,7 +75,6 @@ This can be used to enable minor modes for Python development."
 (defcustom elpy-modules '(elpy-module-sane-defaults
                           elpy-module-company
                           elpy-module-eldoc
-                          elpy-module-autodoc
                           elpy-module-flymake
                           elpy-module-highlight-indentation
                           elpy-module-pyvenv
@@ -3818,11 +3817,8 @@ description."
   "Timer to refresh documentation.")
 
 (defcustom elpy-autodoc-delay .5
-  "The idle delay in seconds until documentation is refreshed automatically.
-
-If nil, deactivate the documentation automatic refresh."
-  :type '(choice (const :tag "never (nil)" nil)
-                 (const :tag "immediate (0)" 0)
+  "The idle delay in seconds until documentation is refreshed automatically."
+  :type '(choice (const :tag "immediate (0)" 0)
                  (number :tag "seconds"))
   :group 'elpy
   :group 'elpy-autodoc)
--- a/elpy.el
+++ b/elpy.el
@@ -75,6 +75,7 @@ This can be used to enable minor modes for Python development."
 (defcustom elpy-modules '(elpy-module-sane-defaults
                           elpy-module-company
                           elpy-module-eldoc
+                          elpy-module-autodoc
                           elpy-module-flymake
                           elpy-module-highlight-indentation
                           elpy-module-pyvenv
@@ -98,6 +99,8 @@ can be inidividually enabled or disabled."
                      elpy-module-yasnippet)
               (const :tag "Django configurations (Elpy-Django)"
                      elpy-module-django)
+              (const :tag "Automatically update documentation (Autodoc)."
+                     elpy-module-autodoc)
               (const :tag "Configure some sane defaults for Emacs"
                      elpy-module-sane-defaults))
   :group 'elpy)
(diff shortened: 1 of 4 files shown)


elpy.el
```diff
@@ -593,9 +593,13 @@ virtualenv.
...
    (elpy-mode
-    (elpy-modules-buffer-init))
+    (elpy-modules-buffer-init)
+    (add-hook 'pre-command-hook 'elpy-autodoc--pre-command nil t)
+    (add-hook 'post-command-hook 'elpy-autodoc--post-command nil t))
    ((not elpy-mode)
-    (elpy-modules-buffer-stop))))
+    (elpy-modules-buffer-stop)
+    (remove-hook 'pre-command-hook 'elpy-autodoc--pre-command t)
+    (remove-hook 'post-command-hook 'elpy-autodoc--post-command t))))
```
jorgenschaefer: Can't this be a module? Then this init/deinit can be moved to a module definition, and people can deactivate it entirely when they want.

Nicholas-Mitchell:
It was me who originally asked for this feature - big thank to @galaunay for the implementation.
I have tested it with my setup in Emacs 25.1.1 with the latest `elpy` from melpa ():

        Status: Installed in ‘elpy-20180209.146/’ (unsigned).
        Version: 20180209.146

I'm happy to report it works really well, as desired, with my current setup. I use the ipython interpreter and so with the latest elpy version, I just had to update my config from the old `(elpy-use-ipython)` to the new cleaner way:

    (setq python-shell-interpreter "ipython"
        python-shell-interpreter-args "-i --simple-prompt")

I didn't have to change anything.

The 0.5 second refresh rate is perfect - no feeling of lag whatsoever, but I don't have the feeling I have to wait on it to read the docs.

galaunay:
@jorgenschaefer I agree with the module. I am working on it.

@Nicholas-Mitchell, thanks for the backup.
Happy to see that It works well and that you find it useful as it is :).

Just an additional piece of information:
Autodoc is not as smart as the classical `elpy-doc`.
For example in this case (`|` being the cursor):
```Python
res = foo(a|, b, c)
```
`elpy-doc` gives the documentation of `foo` while autodoc give the documentation of the type of `a` (which is not very usefull in general...).

This is linked to the fact that I wanted autodoc to work asynchronously.
I don't see any (simple) way of make autodoc smarter...

galaunay:
Done, should be good.

Nicholas-Mitchell:
@galaunay - just an idea about your last point - autodoc returning the documentation for types etc.
Would it be possible to bootstrap something like paredit, to know what the current most relevant sexp is, then pipe that into elpy-doc asynchronously?

Even with this small issue, I am really enjoying autodoc - thanks again!

galaunay:
Good idea.

Currently, when using `elpy-doc`, elpy looks for documentation at points, and if it is not relevant, move backward with `elpy-nav-backward-up-list` and search again for documentation.

To stay asynchronous, we need to know a priori (without calling for documentation) if the current symbol under point is relevant or not.
The only way I could find to do that is to check if the symbol under point ends with an opening parenthesis...

@jorgenschaefer, do you know if python-mode provide a better way of doing this ?

jorgenschaefer:
"Relevant" is a tricky concept :-D I doubt python-mode can do that. We used to have some attempts to do that, and it did not work well.

For this feature, you could set a marker at point, call the backend, and when it returns with no info, continue from the marker instead of point. Just remember to clean up the marker.

Or remember point and continue only if point has not changed, if that works with this workflow. You run a new async call when point changes anyhow, I think.

galaunay:
Ok, 

I removed the parenthesis thing, that was a little subjective (I agree).
After all, somebody may want to access the documentation of builtins type :).

I will use it as it is, see if it really needs a smarter doc function.

galaunay:
@jorgenschaefer do you agree to make this a default behavior ?

jorgenschaefer:
What do you mean with "this", exactly? The whole module?

galaunay:
Sorry for the delay (I somehow missed your reply).

I meant to activate the autodoc module by default, 
which will imply that the documentation will be updated with the thing under point by default.

I can see several reasons for which users may not want this functionality, that's why I am unsure about activating it by default or not.

Nicholas-Mitchell:
Although I love the functionality, I would suggest leaving it off by
default and set a toggle to activate it. The reason being that beginners
may run into problems when documentation isn't found (due to virtual
environment and path issues) or the RPC process times out. In my opinion,
this would negatively effect the entire Elpy out-of-the-box experience.

The documentation should just make it clear, how one activates it.

galaunay:
I updated the documentation and deactivated the module by default.

This should change nothing to the Elpy default behavior, tests are passing and I just checked it on one of my script to ensure that everything was fine.

Seems ready to merge for me.

Thanks all for your help on this.

PR #1290, opened 2018-02-10
Fix django options don't appearing in elpy-config subgroup

galaunay:
## Issues
Django options doesn't appear in `elpy-config` associated subgroup.
(Would appreciate confirmation of this not working for other people...).

## Fix 
Adding Django custom variables to the `elpy-django` customization group allows those variables to appear correctly in `elpy-config`.


gopar:
Hmm the django options show up within `elpy`

Also, the `elpy.el` module contains functions related to Django, they might need to be moved to `elpy-django.el` for it to be consisted and moved to the `elpy-django` group. @jorgenschaefer @galaunay  thoughts?

galaunay:
Weird...
I can't see Django options from `elpy-config`, even with `emacs -Q` (version 25.3.1) and a clean elpy install.
Are you aware of any recent changes in emacs that could have led to this issue ?

@gopar If some functions are only useful for the Django module, I think they belong to `elpy-django.el`.
Are you thinking of the test runners or something else ?

gopar:
I don't know of anything that could of changed. using 25.1.1

Did `emacs -Q` and still see them

screen shot 2018-02-18 at 1 00 38 pm

galaunay:
@gopar I meant they are not appearing in the "Django Extension" subgroup (which is consequently empty...):

This PR just populate the django subgroup:


Always clearer with pictures :)

gopar:
ohhhh, thats what you meant :)

Yeah, i'm for this pr then 👍 Just needs updating to include the `django-test-runner-formats` in `elpy-django` \o/

galaunay:
After a rebase it should do!
I will just wait for the test to pass to merge.

PR #1291, opened 2018-02-11
Allow different Django Test Runners

gopar:
The user can now add test runners into a list if they need to put the correct format
Not sure if this is the right approach but some people do use custom test runners 

Should also probably add this in the docs if it's approved


jorgenschaefer:
Nice! Seems to me the variable should be called `elpy-django-test-runner-formats` or `-separators` or something?

gopar:
Updated, I went with `-formats` suffix.

gopar:
If i get another 👍 i'll merge.

Thanks \o/

jorgenschaefer:
:+1:

Issue #1297, opened 2018-02-13
Error from python-shell-completion-at-point


mambolevis:
Hi,

The following error appears in the `*Python*` buffer when it tries any autocompletion.  
`In [2]: ne`       
After typing one or two words, this error is shown:
```
Company: An error occurred in auto-begin
Company: backend (company-capf :with company-yasnippet) error "Invalid search bound (wrong side of point)" with args (prefix)
```
Any idea how to deal with this issue?
Thanks

galaunay:
Hi, 

I got the same error some months ago.
In my case, it os apparently due to a bug in company, that modify the cursor position before calling `python-shell-completion-at-point`.

A possible workaround that works for me:
```Elisp
(defadvice python-shell-completion-at-point (before fix-company-bug activate)
  "Fix a bug in company, modifying the current point"
  (goto-char (point-max)))
```

For a more sustainable solution, you will have to report this to company and/or python.

mambolevis:
@galaunay 
Yes, the workaround works. Thank you.
I reported it in company-mode. 
What is the official place (repo) to report it in python?

galaunay:
Happy that the workaround works for you too, 
this shows that there is indeed a problem.

Python-mode is part of Emacs, 
so problems with python-mode should be reported directly to emacs.

The best way to do it is by using `report-emacs-bug`.
You can have more infos on how to properly report an issue by doing `report-emacs-bug-info`.

If I were you, I'll wait for an answer from company before reporting to emacs.

mambolevis:
Thank you. 
Ok. I will follow your advice.

galaunay:
Alright, 
Thank you for reporting this.

mambolevis:
I reopen this issue in order to provide a most robus solution according to 
https://github.com/company-mode/company-mode/issues/759
I also changed the titles for a better reference.

galaunay:
Thanks for the feedback,
their solution is way better than mine.
 
I made a PR to be sure this won't bug anyone else.

mambolevis:
Thank you for answering my very first question, that gave me the opportunity not only to learn but also to explore possible solutions and alternatives.

mambolevis:
@galaunay 
Hi, I updated elpy in order to test your fix https://github.com/jorgenschaefer/elpy/pull/1301/commits/da2bfaf387d5b176f33a11905f5894d0e3879668
Two important conclusions:
[...]

PR #1301, opened 2018-02-20
Fix company bug when using completion in the interpreter

galaunay:
Follow issue #1297.

## Issue
`company-sort-by-occurence` does not work in inferior-python-mode and can prevent company from showing up (see company-mode/company-mode#759).

## Solution
As proposed in company-mode/company-mode#759, this PR just make sure `company-sort-by-occurence` is deactivated in `inferior-python-mode`.

This workaround will have to be removed when(if) the issue disappear.


jorgenschaefer:
Looks good to me.

galaunay:
It was not enough for Travis (I missed a parenthesis in fact...).
It is now.
Merging.

PR #1302, opened 2018-02-25
Add django-format to elpy-django group

gopar:


gopar:
Very small change. Ganna merge this in. Just wanted to make sure tests pass

PR #1306, opened 2018-03-03
Add customizeable python shell root

gopar:
So I have started following a certain pattern in python by putting my packages/modules inside a `src` directory (https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure)
Example structure of proj:
```
.
├── README.md
├── setup.cfg
├── setup.py
├── src
│   └── main_package
│       ├── __init__.py
│       ├── some_nested_package
├── tests
│   ├── __init__.py
└── tox.ini
```

This puts me in a weird spot where I do want the shell to have the project root in sys.path but I always want it to start inside of the `src` dir. Any arguments for/against?

### DIFF
--- a/elpy-shell.el
+++ b/elpy-shell.el
@@ -104,6 +104,13 @@ nil, the current directory is used instead."
   :type 'boolean
   :group 'elpy)
 
+(defcustom elpy-shell-custom-root nil
+  "Custom directory where a Python shell will *always* start from.
+
+If you wish to use this, then set `elpy-shell-use-project-root' to nil"
+  :type 'string
+  :group 'elpy)
+
 (defcustom elpy-shell-cell-boundary-regexp
   (concat "^\\(?:"
           "##.*" "\\|"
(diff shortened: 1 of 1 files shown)


galaunay:
In my opinion it makes sense, 
the only argument that I can see against it, is that it's increasing the number of option.

So instead of adding this new option, would it make sense to you to make the option `elpy-shell-use-project-root` accept custom folders ?

It will implies renaming this option to something like `elpy-shell-root-starting-directory` with the possibility to set it to `buffer directory`, `project-root` or `custom directory`.

(It should also avoid conflicts between the two options `elpy-shell-custom-root` and `elpy-shell-use-project-root`)

jorgenschaefer:
Nice idea, and good points from @galaunay.

gopar:
I like the idea. I'll do the change once I get a chance

galaunay:
@gopar Any chance to get a hand on this ?
Do you want me to take over ?

gopar:
Hey @galaunay 
I don't think I'll have time to help on elpy any time soon.
You can take over if you'd like

galaunay:
Hi @gopar, 

No problem, I will just close this PR (which related branch seems to have disappeared ?) and open a new one.

PR #1315, opened 2018-03-16
Source yapf file from project root

galaunay:
Copy of the PR #989 from @jonmoore.

Plus some documentation for the potential users of yapf style file.

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -2101,19 +2101,24 @@ prefix argument is given, prompt for a symbol from the user."
 (defun elpy-yapf-fix-code ()
   "Automatically formats Python code with yapf."
   (interactive)
-  (elpy--fix-code-with-formatter "fix_code_with_yapf"))
+  (elpy--fix-code-with-formatter "fix_code_with_yapf"
+                                 (or (expand-file-name (elpy-project-root))
+                                     default-directory)))
 
 (defun elpy-autopep8-fix-code ()
   "Automatically formats Python code to conform to the PEP 8 style guide."
   (interactive)
   (elpy--fix-code-with-formatter "fix_code"))
 
-(defun elpy--fix-code-with-formatter (method)
+(defun elpy--fix-code-with-formatter (method directory)
   "Common routine for formatting python code."
   (let ((line (line-number-at-pos))
         (col (current-column)))
     (if (use-region-p)
-        (let ((new-block (elpy-rpc method (list (elpy-rpc--region-contents))))
+        (let ((new-block (elpy-rpc method
+                                   (if directory
+                                       (list (elpy-rpc--region-contents) directory)
+                                     (list (elpy-rpc--region-contents)))))
               (beg (region-beginning)) (end (region-end)))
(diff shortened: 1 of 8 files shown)


jorgenschaefer:
Looks good to me. I guess you can just always pass the directory there, reducing a lot of conditionals, don't you think?

galaunay:
It appears that `elpy-autopep8-fix-code` was suffering the same problem:
configurations files in the project root was not taken into account.

It is now fixed.

It also allowed to reduce the complexity of `elpy--fix-code-with-formatter` (less conditionals :) )

elpy.el
```diff
@@ -2099,30 +2099,43 @@ prefix argument is given, prompt for a symbol from the user."
...
+        (directory (or (expand-file-name (elpy-project-root))
+                              default-directory)))
     (if (use-region-p)
-        (let ((new-block (elpy-rpc method (list (elpy-rpc--region-contents))))
-              (beg (region-beginning)) (end (region-end)))
+        (let ((new-block (elpy-rpc method
+                                   (list (elpy-rpc--region-contents)
+                                         directory)))
+              (beg (region-beginning))
+              (end (region-end)))
```
jorgenschaefer: Much nicer layout here, thank you. The old one has annoyed me for a while.

elpy.el
```diff
@@ -2099,30 +2099,43 @@ prefix argument is given, prompt for a symbol from the user."
...
            beg end
            (replace-regexp-in-string "\n$" "" new-block))
           (goto-char end)
           (deactivate-mark))
       ;; Vector instead of list, json.el in Emacs 24.3 and before
       ;; breaks for single-element lists of alists.
-      (let ((new-block (elpy-rpc method (vector (elpy-rpc--buffer-contents))))
+      (let ((new-block (elpy-rpc method
+                                 (list (elpy-rpc--buffer-contents)
+                                       directory)))
```
jorgenschaefer: The comment can go away, right?

elpy/auto_pep8.py
```diff
@@ -11,11 +12,15 @@
...
 
     """
     if not autopep8:
         raise Fault('autopep8 not installed, cannot fix code.',
                     code=400)
-    return autopep8.fix_code(code, apply_config=True)
+    old_dir = os.getcwd()
+    os.chdir(directory)
+    ret = autopep8.fix_code(code, apply_config=True)
+    os.chdir(old_dir)
```
jorgenschaefer: Use `try`/`finally` here to ensure the directory is changed back even in case of an error.

elpy.el
galaunay: didn't noticed it,
It is removed now.

elpy/auto_pep8.py
galaunay: Done

jorgenschaefer:
Sans the broken tests, this looks good to me! Thank you!

galaunay:
Tests are up to date,
merging !

Issue #1200, opened 2017-10-23
elpy-test uses wrong syntax to invoke Django runner with specific test


twhaples:
I have a new installation of emacs-prelude with `virtualenvwrapper.el` and `elpy`. I am trying to run Django tests and have configured the Django testrunner with

```
(setq elpy-test-django-runner-command '("./manage.py" "test"))
```

Running `elpy-test` while the point is in whitespace at the top of test file works just fine, running all tests in the file, but when the point is inside a specific test example, running `elpy-test` fails:
```
./manage.py test org.module.blah.tests.test_org_blah_api.BlahApiTestCase.test_blahness
...
ERROR: Failure: ImportError (No module named test_blahness)
```

This command also fails on the command line. It is my understanding that this should instead be
```
./manage.py test org.module.blah.tests.test_org_blah_api:BlahApiTestCase.test_blahness
```

with a colon between the filename and the class. This actually works from the command line.

This is with `django-nose==1.4.5` and `Django==1.11.6`.

jorgenschaefer:
Huh! Thank you for the report. Did this change recently in Django?

twhaples:
I really have no idea; I literally just got to this ecosystem last week

jorgenschaefer:
Heh. Welcome, and enjoy your stay. :-)

gopar:
Looks like this has always been the case with `nose`

michaelsteigman:
Having similar trouble with `django-nose==1.4.4` and `Django==1.10.5` and latest elpy. In my case, `elpy-test` inside of an individual test case results in "Ran 0 tests in 0.000s", which is consistent with what I see on the command line using dotted path with no colon before test class.

gopar:
Yeah, the issue here is that `elpy-test-django-runner-manage-command` thinks that these test are run the traditional way, but nose is what actually runs when `manage.py test` is called. I suppose a way to solve this is to decide what format to call the tests depending on what the value of `elpy-test-runner`. That's the quickest solution that I can come up with. Any other ideas?

jorgenschaefer:
Don't the test runners themselves build that path? E.g. `elpy-test-django-runner` at line 1980?

gopar:
Note for myself: There is a way to set the testrunner via `manage.py`. Can probably use this to correct behavior, similar to how `elpy-test-runner` works but maybe call it `elpy-django-test-runner`?

[...]

PR #1317, opened 2018-03-18
Add missing function back

gopar:
Looks like somewhere along the way, this function was removed :P 

Related to #1200


jorgenschaefer:
Oops :-D Thank you!

Issue #1308, opened 2018-03-04
15 failing tests in what will be the official Debian package


sten0:
Hi Jorgen,

Thank you for elpy!  A number of my colleagues are excited about its official inclusion into the Debian (and Ubuntu via Debian) archive.  As part of its packaging I've been working on making the self-tests succeed using Debian's build environment and policy.  eg: using only Debian-packaged dependencies (no network access, and no pip); using virtualenv seems to be against policy, because the autotests must be against the package as-installed, using only system libraries; finally, we cannot use Cask.

This said, I've successfully configured my package such that 329/344 tests pass using Debian's ert-testing framwork (dh_elpa_test)!  At the moment I've disabled the following failing tests, because 329/344 is better than no testing at all (a build fails if any test fails).

Ran 344 tests, 329 results as expected, 15 unexpected (2018-03-03 08:25:59+0000)

15 unexpected results:
   FAILED  elpy-config--insert-configuration-problems-should-not-fail
   FAILED  elpy-config-should-not-fail
   FAILED  elpy-config-should-not-fail-without-virtualenv
   FAILED  elpy-config-should-show-flake8-pip-button-when-no-syntax-checker-available
   FAILED  elpy-insert--pip-button-value-create-should-do-user-install
   FAILED  elpy-rpc-integration
   FAILED  elpy-shell-kill-all-should-kill-all-shells
   FAILED  elpy-shell-send-current-statement-should-send-current-statement-only
   FAILED  elpy-shell-send-region-or-buffer-should-notify-of-removing-main
   FAILED  elpy-xref--apropos-should-give-an-xref-buffer
   FAILED  elpy-xref--definitions-should-return-definitions
   FAILED  elpy-xref--get-completion-table-should-return-completion-table
   FAILED  elpy-xref--references-test
   FAILED  elpy-yapf-fix-code-in-region-should-retain-line-and-column
   FAILED  elpy-yapf-fix-code-should-retain-line-and-column

If you have the time, and if these tests are important enough to be worth your time, then would you please consider helping to make them work?  Some, like `elpy-config-should-not-fail-without-virtualenv` and `elpy-config-should-show-flake8-pip-button-when-no-syntax-checker-available` can just be disabled, for the purposes of Debian, but I'm concerned about many of the others. 

[...]

PR #1318, opened 2018-03-20
Use hungry glob for string-match in test/elpy-config-test.el

sten0:
Re: #1308 

This did the trick.  Filing a pull request in case it's faster to click and "Ok, merge this" button.


jorgenschaefer:
Yes it is :-D

PR #1327, opened 2018-03-26
Bug fix: elpy-shell-send-file was not interactive. 

sam-s:
elisp is not python - triple quotes are not okay.


galaunay:
You are definitely right, 
(and about the rest as well).

Thanks for this.

PR #1329, opened 2018-03-30
Add github PR and issue templates

galaunay:
There is a significant potion of the issues that take times to resolve due to lack of information.
In the hope that this will make the process easier, I added an issue template to indicate to the users what kind of information we need.

I also added a PR template, to be sure no one forget about tests, documentation updates and changelog updates (which I forget myself quite often...).

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,11 +1,10 @@
 # Summary
 
+Thank you for reporting to Elpy !
 
-# Expected behavior
-
-
-# Actual behavior
-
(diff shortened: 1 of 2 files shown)


jorgenschaefer:
I like it. I hope it does not dissuade people from filing bugs, but anything that makes life easier for us is great IMHO …

gopar:
completely for it.

galaunay:
@jorgenschaefer @gopar 
Does it seems more appealing with a small explanation at the beginning ?

PR #1330, opened 2018-04-01
Suggest elpy-rpc-restart after setting PATH

offby1:
# PR Summary

When warning about `~/.local/bin` not being on `PATH`, suggest running `elpy-rpc-restart` to make changes to `PATH` "stick".

I'm skipping the various "pull request" checkboxes because this change is so small.


gopar:
Thanks!

PR #1332, opened 2018-04-02
Modify UTF8 character "ö" in elpy-rpc-test to allow test to pass

sten0:
This might be caused by a UTF8 handling bug in Debian's dh-python,
dh-elpa, libdpkg-perl, or a limitation in pbuilder; however, this
simple patch allows the test to pass when building in a pbuilder or on
the buildd network.

This patch fixes one of the failing tests discussed in #1308 .  Previously I was disabling the test, so I consider it a win!  I apologise in advance for my limited knowledge of Germanic languages, because I've probably mistransliterated "ö" as "o".

### DIFF
--- a/elpy/tests/test_server.py
+++ b/elpy/tests/test_server.py
@@ -332,11 +332,11 @@ class TestGetSource(unittest.TestCase):
         fd, filename = tempfile.mkstemp(prefix="elpy-test-")
         self.addCleanup(os.remove, filename)
         with open(filename, "wb") as f:
-            f.write(u"möp".encode("utf-8"))
+            f.write(u"mop".encode("utf-8"))
 
         source = server.get_source({'filename': filename})
 
-        self.assertEqual(source, u"möp")
+        self.assertEqual(source, u"mop")
 
 
 class TestPysymbolKey(BackendTestCase):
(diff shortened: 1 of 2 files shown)


sten0:
Please feel free to reject this patch if you'd like.  I'm including it in the Debian package so that I can provide a test case for my bug report.  Honestly I don't know if it is useful for others...

sten0:
Hi Jorgen,

Chris Lamb raised concerns about my patch, eg: that if UTF-8 characters aren't being tested, then elpy-rpc-test might fail during actual usage.  If it's a known bug in elpy then please cancel this pull request, and I'll disable the test in Debian.  Otherwise I'll continue continue exploring the hypothesis that the bug is in Debian.

@lamby

lamby:
This test seems *deliberately* designed to find UTF-8 errors, so to patch it out because you are seeing problems doesn't feel right to me :-)

jorgenschaefer:
Hello! This test is indeed meant to catch encoding errors. This regularly was a problem on certain platforms with non-standard encodings, including Windows. As this works on our CI and on multiple machines, the hypothesis that this is broken in the Debian build environment seems rather plausible. Try setting the `LC_*` variables correctly, for example.

Issue #1333, opened 2018-04-03
elpy-config says I'm using the python from virtualenv but it is not.


AkashGanesan:
# Summary

When I use pyvenv-activate, it activates the environment but it doesn't use the virtualenv's python even though elpy-config says so.  Also, I don't have exec-shell-from-path 

When I try the following after activating the environment, 
````
import sys
print sys.executable
````
It prints the global path instead of the one in the ENV.
# My configuration
## OS
Ubuntu 17.04

## Result of `(elpy-config)`
Virtualenv........: ENV (/home/akash/learn/504/final_project/EECS_504_Project/keras_Realtime_Multi-Person_Pose_Estimation/ENV)
RPC Python........: 3.6.3 (/home/akash/learn/504/final_project/EECS_504_Project/keras_Realtime_Multi-Person_Pose_Estimation/ENV/bin/python)
Interactive Python: jupyter (/home/akash/learn/504/final_project/EECS_504_Project/keras_Realtime_Multi-Person_Pose_Estimation/ENV/bin/jupyter)
Emacs.............: 25.2.2
Elpy..............: 1.19.0
Jedi..............: 0.11.1
Rope..............: Not found (0.10.7 available)
Autopep8..........: 1.3.5
Yapf..............: 0.21.0
Syntax checker....: flake8 (/home/akash/learn/504/final_project/EECS_504_Project/keras_Realtime_Multi-Person_Pose_Estimation/ENV/bin/flake8)

## Elpy configuration in my init.el
```
(elpy-enable)
(setq python-shell-interpreter "jupyter"
      python-shell-interpreter-args "console --simple-prompt")
;; To prevent exec-from-shell :: https://github.com/jorgenschaefer/elpy/issues/1001
(setq shell-file-name "/bin/bash")
```

**Comment by galaunay** [comment]
Thanks for reporting this.

I was not able to reproduce your issue.
Have you restarted the python shell after changing the environment ?
with `elpy-shell-kill-shell` and `elpy-shell-switch-to-shell` for example ?

**Comment by AkashGanesan** [comment]
Hi @galaunay :  I figured out why this went wrong.  In the virtualenv, jupyter was not installed.  When I installed this, it works.  However, without juypter, I get the above when I do elpy-config though.  Maybe elpy can tell when there's no virtualenv based jupyter.

**Comment by galaunay** [comment]
Thank you for the feedback.

Elpy is supposed to fall back to the system installation if jupyter is not installed in the virtualenv:
```
[...]

PR #1335, opened 2018-04-04
Raise an error if the interpreter is not in the current virtualenv

galaunay:
# PR Summary
Following issue #1333.
Elpy now raises an error if the wanted interpreter (set by `python-shell-interpreter`) is not available within the current virtual environnement.
This avoid Elpy to just fall back to the system interpreter binaries, outside of the virtual environment, without the user noticing.

# PR checklist

### DIFF
--- a/elpy-shell.el
+++ b/elpy-shell.el
@@ -229,6 +229,11 @@ Python process. This allows the process to start up."
          (proc (get-buffer-process bufname)))
     (if proc
         proc
+      (when (and pyvenv-virtual-env
+                 (not (string-prefix-p (expand-file-name pyvenv-virtual-env)
+                                       (executable-find
+                                        python-shell-interpreter))))
+        (error "Intepreter binaries not found in current virtual environnement"))
       (let ((default-directory (or (and elpy-shell-use-project-root
                                         (elpy-project-root))
                                    default-directory)))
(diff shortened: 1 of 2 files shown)


elpy-shell.el
```diff
@@ -229,6 +229,11 @@ Python process. This allows the process to start up."
          (proc (get-buffer-process bufname)))
     (if proc
         proc
+      (when (and pyvenv-virtual-env
+                 (not (string-match (regexp-quote pyvenv-virtual-env)
```
jorgenschaefer: Should this use `string-prefix-p`, and maybe `expand-file-name` for both?

galaunay:
I also added some tests.
@jorgenschaefer Do you know if there is a way to create a temporary virtual environment to make this test more efficient ?

jorgenschaefer:
> @jorgenschaefer Do you know if there is a way to create a temporary virtual environment to make this test more efficient ?

Creating a "real" virtualenv is a tricky process that depends on the Python version etc., so I don't think it's something we can do easily. virtualenvwrapper has a temporary virtualenv feature, but that just adds a post-deactivate hook that deletes the directory. You can either create a virtualenv for the test, or you can just set `exec-path` to something in a temporary directory and fake it that way … activating a virtualenv really is just changing `exec-path`.

galaunay:
Ok, thanks for the clarification.
In this case I guess it is good enough to just activate an empty virtual env.

PR #1342, opened 2018-04-15
Avoid-test-error-due-to-rpc-timeout

galaunay:
# PR Summary
`elpy-xref--references-test` have been producing random test errors in Travis (in #1341 at least).
As those errors only appear on Travis, this is a test PR to investigate this issue.

My first guess is that those errors are due to rpc timeout.
I increased the rpc timeout to see if travis is still failing or not.

# PR checklist


jorgenschaefer:
Good guess :-)

galaunay:
Now another test (`elpy-doc-should-find-documentation-from-inside-arguments`) is failing because of timeout.

Should we set `elpy-rpc-timeout` to an arbitrary big value for all tests ?

jorgenschaefer:
Probably a good idea, if Travis has speed issues.

galaunay:
Okay...
There was something else going wrong.

`elpy-xref--references-test` was failing for emacs 25.
It was not the last time the test was modified (#1258).

No idea why, but a simple `(save-buffer)` was enough to fix it.

Issue #1246, opened 2017-12-16
Bug in snippet __init__ when including type annotations


gopar:
Currently if you type `__init__` in a buffer you get a nice little snippet that fills up the instance variables with the arguments your passing:

```python
def __init__(self, a, b):
    self.a = a
    self.b = b
```

But if you include type annotations, it gets all weird:

```python
def __init__(self, a: int, b: str=None):
    self.a: int = a: int
    self.b: str = b: str
```

The regex from the `elpy-snippet-split-args` function seems to be the problem.

I tried the following regex on regex101.com and worked but failed when I tried in emacs: 
`[[:blank:]]*:?[\w+]*=[[:blank:]]*`

galaunay:
I had more success with this one:
```Elisp
(defun elpy-snippet-split-args (arg-string)
  "Split a python argument string into ((name, default)..) tuples"
  (mapcar (lambda (x)
            (split-string x "[[:blank:]]*\\(:[[:blank:]]*[[:alpha:]]+\\)?[[:blank:]]*=?[[:blank:]]*" t))
          (split-string arg-string "[[:blank:]]*,[[:blank:]]*" t)))
```
But maybe there is a more simple one to do the same thing...

gopar:
Thanks, my regex fu is beyond horrible. That does seem rather long though :P

jorgenschaefer:
`rx` is also very convenient to construct complex regular expressions. These might be useful to place into global variables, too.

PR #1343, opened 2018-04-15
Make snippets able to handle arguments with type annotations.

galaunay:
# PR Summary
Follow discussion in #1246.

Make snippets (notably `__init__`) able to handle arguments with type annotations.

`elpy-snippet-split-args` now uses regular expressions to extract argument name (instead of using `split-string`).

# PR checklist


galaunay:
Added tests to be sure everything works fine.

jorgenschaefer:
I like it. Does this interact with the attempts to merge the snippets into yasnippets? https://github.com/AndreaCrotti/yasnippet-snippets/pull/259

galaunay:
Yes it does :).

@sten0 do you think this can be integrated in the yasnippets PR ?

sten0:
@galaunay 1ad8dc56f2a2ee37cbbd7b421e7a380fa8cd4dda definitely can be.  If you want to add it to the yasnippets PR and want all the glory, then please fork my fork, apply your patch to the tip of the merge-elpy-snippets-WIP branch, and submit a pull request.  https://github.com/sten0/yasnippet-snippets/tree/merge-elpy-snippets-WIP  I'd prefer not to do this work myself ;-)  but I'd be happy to if you insist.

Do the tests actually depend on elpy-mode?  If not then I think it would be really cool to add them to yasnippet-snippets, because I think your work could inspire others to write more powerful snipppets and self-tests.  Yasnippet-snippets doesn't have any self-tests of its own yet, so I see it as an opportunity to lead by example.

gopar:
Looks like the merge for yasnippets didn't happen? :(

galaunay:
This has been delayed because all the snippets were supposed to be merged into yasnippet.
Until they are indeed merged, they will be available in Elpy.

sten0:
@galaunay, thank you for your work on this, and sorry I forgot about it.  165c72d will make the next attempt much faster.  At the moment I'm focused on getting through my back-queue before the upcoming Debian freeze, plus bumping the urgency for QA work.

Issue #1345, opened 2018-04-17
C-c C-R f to to fix code throws wrong-type-argument on Emacs 26.1


ViktorHaag:
# Summary

With Emacs 26.1, and package-installed elpy, using elpy to re-format code throws `wrong-type argument stringp nil`.

# Steps to reproduce
- Start emacs
- open a simple python script 
- C-c C-R f after a long line

## Debugger trace
```
Debugger entered--Lisp error: (wrong-type-argument stringp nil)
  expand-file-name(nil)
  elpy--fix-code-with-formatter("fix_code_with_yapf")
  elpy-yapf-fix-code()
  elpy-format-code()
  funcall-interactively(elpy-format-code)
  call-interactively(elpy-format-code nil nil)
  command-execute(elpy-format-code)
```

# My configuration
## OS
macOS 10.12.6

## Result of `(elpy-config)`
```
Virtualenv........: None
RPC Python........: 3.6.5 (/usr/local/bin/python3)
Interactive Python: python3 (/usr/local/bin/python3)
Emacs.............: 26.1
Elpy..............: 1.19.0
Jedi..............: 0.12.0
Rope..............: 0.10.7
Autopep8..........: 1.3.5
Yapf..............: 0.21.0
Syntax checker....: flake8 (/usr/local/bin/flake8)
```

## Elpy configuration in my init.el
```
;;; Elpy mode for python
;;; uses starter-kit funcs to install packages if needed/not yet installed
(starter-kit-install-if-needed 'elpy 'flycheck 'py-autopep8 'py-yapf)
(elpy-enable)
(setq elpy-rpc-python-command "python3")
(setq elpy-modules '(elpy-module-company
                     elpy-module-eldoc
                     elpy-module-pyvenv
                     elpy-module-yasnippet
                     elpy-module-sane-defaults))
(require 'py-yapf)
(setq elpy-rpc-timeout 10)
```

galaunay:
Thanks for the report.

I was able to reproduce the bug.
For me It happens when the script is not in a project and `(elpy-project-root)` is nil.

I made a PR (#1346) to fix this.

PR #1346, opened 2018-04-17
Prevent 'expand-file-name' from erroring when project-root is nil

galaunay:
# PR Summary
Fix issue #1345:
`elpy-format-code` raise an error when the script is outside a project.

# PR checklist


ViktorHaag:
Confirmed that this patch addresses me issue in my local install.

galaunay:
Thanks for the feedback.
I am merging this, then.

Issue #1328, opened 2018-03-27
The shell buffer and "elpy-shell-toggle-dedicated-shell"


Nicholas-Mitchell:
Thank you for this package, specifically for a recent enhancement in the latest update; namely the ability to set a buffer to run into a specific shell. Here is the related [enhancement issue - [#1232] ](https://github.com/jorgenschaefer/elpy/pull/1232).

I am having a few problems with this feature. I can assign buffers their own shells and send code to them as one would expect - works great!

## The scenario:
If I use the function `elpy-shell-toggle-dedicated-shell` and select a different shell, there is no error or warning, however things don't work as expected. If I send a line of code from the buffer to the newly assigned shell, the literal line of code does appear in the new shell, however it is actually executed in the original shell.

The iPython interpreter of the original shell does update (the `In [n]` is incremented to `In [n+1], but there is otherwise no text shown`), whereas the newly assigned shell just gets the raw text and the interpreter seems to hang, e.g.:
```python
In [6]: import matplotlib.pyplot as plt
test_loss = -np.inf
BATCH_SIZE_TRAIN = 512
BATCH_SIZE_VAL = 256
```
... this only updates to have `In [7]` if I go into the shell and hit <kbd>Enter</kbd>.
The values e.g. `test_loss = -np.inf` are not defined in that shell, i.e. they didn't run.

If I assign the buffer back to the original shell, things carry on working as expected. this behaviour occurs if I use either of the new functions to swap the shell assignment, i.e. `elpy-shell-toggle-dedicated-shell` or `elpy-shell-set-local-shell`.

Could this be linked to the way the iPython interpreter works within Emacs?

## My setup:
- elpy 1.18 (both Emacs and Python versions)
- emacs version: 25.3.2 on Linux
- working in a virtual environment, using `pyvenv-workon`
`elpy-config` output:

> Elpy Configuration
> 
> Virtualenv........: sigra-prep (/home/nicholas/anaconda3/envs/env1)
> RPC Python........: 3.5.2 (/home/nicholas/anaconda3/envs/env1/bin/python3)
> Interactive Python: ipython (/home/nicholas/anaconda3/envs/env1/bin/ipython)
> Emacs.............: 25.3.2
> Elpy..............: 1.18.0
> Jedi..............: 0.11.0 (0.11.1 available)
> Rope..............: Not found (0.10.7 available)
> Autopep8..........: 1.3.3 (1.3.4 available)
[...]

PR #1353, opened 2018-04-20
Use new convention for python shell name

galaunay:
# PR Summary

The name of the dedicated python shells are now using the buffer name
without extension, for example `*Python[test]*` instead of `*Python[test.py]*`.

This avoid conflicts with the way python.el handles dedicated shells
(see #1328).

# PR checklist


galaunay:
Not a fundamental modification, I am merging this.

Issue #788, opened 2016-01-30
company module is too intrusive


vspinu:
I think company module is way too intrusive. I have opened an [issue in company](https://github.com/company-mode/company-mode/issues/463) but I guess it's more of an elpy problem. 

Company mode is so ubiquitous that you would expect most users would have their own configuration by now. So sneakly [setting](https://github.com/jorgenschaefer/elpy/blob/master/elpy.el#L3169-L3179)  keys in company map (C-d), messing with dabbrev and setting idle company delay behind the user's back doesn't seem right. Most of the users would have already customized these features themselves.

I think the readme should mentioned `elpy-modules`` and what they overwrite in the backgraound. 

There could even be two versions of `elpy-modules` overwrites - one lightweight and non-contentious which would simply initialize the functionality with most basic defaults, and another one - feature reach and intrusive, like it is right now.

jorgenschaefer:
Hello, and thanks for raising this issue!

Elpy is meant as a mode you can install on a plain Emacs and have it be useful. A lot of the things in Elpy could, arguably, be left out because experienced Emacs users surely can (or already have) configured them themselves.

Elpy goes the other way. We provide a default config that works for people with nothing else installed, and allow experienced users like you to tell it to not configure things you already do yourself. I find it a lot more sensible to expect experienced users to put in some configuration effort than new users.

I'm fine if you want to add some documentation on the wiki that mentions the modules and what they do, of course. Maybe a new page "Elpy for experienced users"? The README is a bit too crowded already for this kind of information, though.

I hope this answers your concerns?

vspinu:
> Elpy goes the other way. We provide a default config that works for people with nothing else installed,

Thanks for this. It's very helpful indeed. The question is, do you want to help new emacs users or  new python users? I think most of elpy "new" users will be of the second  type. 

I guess my main problem is that I cannot easily isolate things what I want to keep. Everything seems to be too tightly coupled. For example, I want to keep company module because of semantic/capf cleanup. But I don't want it to interfere with my keys.  

Elpy rebinds C-d not only in python mode but everywere else. A lot of modes have full doc maps on C-d. Given that company is very often active that completely messes things up.  

> I find it a lot more sensible to expect experienced users to put in some configuration effort than new users.

This is not that it easy as it sound. How do I actually do that? Instalation instructions say very little about that.

[...]

Issue #486, opened 2015-02-08
Failing test: elpy-flymake--standard-value-should-get-standard-value in Emacs 25


birkenfeld:
Since Emacs revision bd7a1e156, `python.el` uses `(executable-find "pyflakes")` for determining `python-check-command`, so that the default value now includes the full path to pyflakes (for me it happens to be in `~/.local/bin`).

`(should (equal (elpy-flymake--standard-value 'python-check-command) "pyflakes"))` now fails.

jorgenschaefer:
Fixed in 24f3d5b

PR #1358, opened 2018-04-25
Stop overriding company customizable variables

vsedach:
There are two problems with the removed code:

1. company-dabbrev-code-everywhere override is not restored.

2. Overriding user customizations for third party libraries with arbitrary values is extremely rude. For example I like to have company-idle-delay set to nil.

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -3309,15 +3309,6 @@ If you need your modeline, you can set the variable `elpy-remove-modeline-lighte
                                       (remove 'company-sort-by-occurrence
                                               company-transformers)))))
     (`buffer-init
-     ;; We want immediate completions from company.
-     (set (make-local-variable 'company-idle-delay)
-          0.01)
-     ;; And annotations should be right-aligned.
-     (set (make-local-variable 'company-tooltip-align-annotations)
-          t)
-     ;; Also, dabbrev in comments and strings is nice.
-     (set (make-local-variable 'company-dabbrev-code-everywhere)
-          t)
      ;; Add our own backend and remove a bunch of backends that
      ;; interfere in Python mode.
      (set (make-local-variable 'company-backends)
@@ -3334,8 +3325,6 @@ If you need your modeline, you can set the variable `elpy-remove-modeline-lighte
         (buffer-name) elpy-rpc-ignored-buffer-size)))
     (`buffer-stop
      (company-mode -1)
-     (kill-local-variable 'company-idle-delay)
-     (kill-local-variable 'company-tooltip-align-annotations)
      (kill-local-variable 'company-backends))
     ))
 
--- a/elpy.el
+++ b/elpy.el
@@ -3294,6 +3294,17 @@ If you need your modeline, you can set the variable `elpy-remove-modeline-lighte
 ;;;;;;;;;;;;;;;;;;;
 ;;; Module: Company
 
+(defvar elpy--company-default-overrides ())
+
+(defun elpy--override-company-default (variable override-value)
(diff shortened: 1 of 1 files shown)


galaunay:
I do agree on the fact that `company-dabbrev-code-everywhere` should be restored.

I also agree that it is rude to override use customization, but I do like the fact that elpy provide a default configuration for company.

I wonder if their is a way of changing the default values for company properties (using `setq-default` ?) but only for buffers in `elpy-mode`...

vsedach:
> I wonder if their is a way of changing the default values for
> company properties (using `setq-default` ?) but only for buffers in
> `elpy-mode`...

The buffer-local values are probably the right way to do that. You can
tell that a customizable variable was changed from the default with an
expression like:

(equal (eval (car (get 'company-idle-delay 'standard-value)))
       company-idle-delay)

This is basically what describe-variable in help-fns.el.gz that comes
with Emacs does (see also 14.3 Defining Customization Variables in the
Elisp manual).

When the value for the variable is unchanged from the default, then I
think it would be safe to bind it buffer-local. This would keep the
existing behavior for users that do not know about these variables,
and stop overriding them for people that do.

Vladimir

vsedach:
> You can easily prevent Elpy from overriding Company settings by
> disabling Elpy's company module in `elpy-modules`, this allows you
> to to only what you want to happen. You can even do `(setq
> elpy-modules nil)` and do everything by hand.

That is not true. elpy-mode-map bindings for those modules would still
be there, autodoc depends on company, etc. The current elpy code is
not modular in the way you describe, and it would take a fair amount
of work to make it that way.

> The only thing that needs changing here is to restore `company-dabbrev-code-everywhere`, apparently that was forgotten when that configuration was added. :-)

I will make a patch that uses the standard-value property to only
shadow defaults, and will automatically manage the buffer-local
shadowing so that omissions like company-dabbrev-code-everywhere do
not happen again, on this PR in the next couple of days.

Thank you,
Vladimir

jorgenschaefer:
> That is not true. elpy-mode-map bindings for those modules would still be there, autodoc depends on company, etc. The current elpy code is not modular in the way you describe, and it would take a fair amount of work to make it that way.

If you feel that Elpy still does too much when `elpy-modules` is nil, you can easily not run `elpy-enable` at all and just use exactly the parts of Elpy you want. You do not even have to use `elpy-mode`!

> I will make a patch that uses the standard-value property to only shadow defaults, and will automatically manage the buffer-local shadowing so that omissions like company-dabbrev-code-everywhere do not happen again, on this PR in the next couple of days.

We tried to do this with some other variables. This approach sadly failed in various ways for users, so we took that functionality out again.

The only PR here that I will accept is one that revers the local variable state for `company-dabbrev-code-everywhere`.

vsedach:
> We tried to do this with some other variables. This approach sadly
> failed in various ways for users, so we took that functionality out
> again.

Can you please point to the commits where you claimed to try this
approach, and the issues encountered? I am having a hard time
believing this claim given that standard-value is used all over Emacs.

Vladimir

vsedach:
> We tried to do this with some other variables. This approach sadly failed in various ways for users, so we took that functionality out again.

Looking more into this issue, I found why you thought this approach
"failed":

https://github.com/jorgenschaefer/elpy/issues/486

Hard-coding values into unit tests that depend on combinations of
arbitrary versions of 3rd party libraries running on Emacs 24 is not
something I want to waste my time on.

Between that and reading the following:

https://github.com/jorgenschaefer/elpy/issues/788

I really do not feel it is a productive use of time to try to work
with you to upstream changes.

PR #1360, opened 2018-04-27
update elpy-flymake-error-at-point for supporting the newly designed …

zhangda82:
…flymake API starting from Emacs 26

added testing of the new mechanism and kept the old mechanism

# PR Summary

# PR checklist

- [ ] Code is not generating new bytecode warnings

## For new features only:

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -3702,7 +3702,11 @@ description."
       (when err-info
         (mapconcat #'flymake-ler-text
                    err-info
-                   ", ")))))
+                   ", "))))
+  (when (and (fboundp 'flymake-diagnostics) (fboundp 'flymake-diagnostic-text))
+    (mapconcat #'flymake-diagnostic-text (flymake-diagnostics (point)) "\n")
+    )
+  )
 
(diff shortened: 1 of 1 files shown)


elpy.el
```diff
@@ -3702,7 +3702,11 @@ description."
       (when err-info
         (mapconcat #'flymake-ler-text
                    err-info
-                   ", ")))))
+                   ", "))))
+  (when (and (fboundp 'flymake-diagnostics) (fboundp 'flymake-diagnostic-text))
+    (mapconcat #'flymake-diagnostic-text (flymake-diagnostics (point)) "\n")
+    )
+  )
```
jorgenschaefer: This whole function probably should use a `cond` form. Also, it should return a newline-separated or comma-separated string, not one or the other depending on the flymake version.

Oh, and in Lisp, closing parens are never alone on a line, they get lonely and scared, and we do not like scared little parens!

Otherwise, looks good to me :-)

jorgenschaefer:
I think this was just fixed by a different PR. Thank you for your work!

zhangda82:
Thanks!

Issue #1357, opened 2018-04-25
elpy still uses "flymake-err-info", which is disabled in Emacs 26.1, in error message reporting.


zhangda82:
# Summary

Hi there. 
I tried to call elpy-flymake-show-error on an error/warning annotated by flymake, but nothing showed up. 
After checking the code of elpy-flymake-show-error and the elpy-flymake-error-at-point, I noticed that 
elpy is still assuming "flymake-err-info" to be the error message. 

> (defun elpy-flymake-error-at-point ()
  "Return the flymake error at point, or nil if there is none."
  (when (boundp 'flymake-err-info)
    (let* ((lineno (line-number-at-pos))
           (err-info (car (flymake-find-err-info flymake-err-info
                                                 lineno))))
      (when err-info
        (mapconcat #'flymake-ler-text
                   err-info
                   ", ")))))

However, in Emacs 26.1, flymake-err-info is no longer available. 
I think this is a compatibility issue. 

Thanks.

Da

zhangda82:
Here is the email communication I had with the maintainer of flymake. The old API needs to be replaced with their new designs:

> Hello --
>
> I am a long term Emacs user and I use flymake together with flymake-cursor on a daily basis.
>
> Starting Emacs 26.1, I noticed that flyamke was redesigned and no
> longer has the var flymake-err-info that contains the error info
> reported by the backend syntax checker.
>
> This makes flymake-cursor and code like the following not work:
>
> (defun elpy-flymake-error-at-point ()
>   "Return the flymake error at point, or nil if there is none."
>   (when (boundp 'flymake-err-info)
>     (let* ((lineno (line-number-at-pos))
>            (err-info (car (flymake-find-err-info flymake-err-info
>                                                  lineno))))
>       (when err-info
>         (mapconcat #'flymake-ler-text
>                    err-info
>                    ", ")))))
>
> I believe this is a potential issue and wonder if someone could direct
> me to any alternatives to flymake-err-info in the newly designed
> flyamke.

Hi -- 
I'm the responsible for the redesign. Although I tried to keep *some*
backward compatibility to old flymake.el's API, the truth is that *all*
of it was an API.

The bad news is that this variable is gone: it couldn't be
backward-adapted. The good news is that it should be much easier to
write your function:

(defun elpy-flymake-error-at-point ()
  "Return the flymake error at point, or nil if there is none."
  (mapconcat #'flymake-diagnostic-text (flymake-diagnostics (point)) "\n"))

[...]

Issue #1272, opened 2018-01-20
Possible future problem with elpy-flymake-next-error


tomnor:
Hello Elpy!

Running Emacs master I have noticed that I am getting no output in the mini-buffer after C-c C-n (elpy-flymake-next-error). Trying to trace the function call a bit, I see that the function `elpy-flymake-error-at-point` starts with `(when (boundp 'flymake-err-info) ...`. But then I could not find that variable in the flymake source code. Also when checking out emacs-26, I could not see it. Checking out emacs-25 I can find it, and by loading flymake.el from that branch and re-visit my python file, I get the output back.

So I was wondering if somebody already noticed this or am i just totally confused. If not totally confused there should be some updates for compatibility with emacs-26.

What do you think?

jorgenschaefer:
Nice catch, thank you. Yep, that should get fixed. (Though I think the "correct" fix is to just switch to flycheck entirely)

tomnor:
> Nice catch, thank you. Yep, that should get fixed. (Though I think the "correct" fix is to just switch to flycheck entirely)

Is that a recommendation to me as a user or a plan for the elpy project?
Can I use flycheck as a drop-in replacement already?

jorgenschaefer:
Both, really. And yes, you can:

```
(when (require 'flycheck nil t)
  (setq elpy-modules (delq 'elpy-module-flymake elpy-modules))
  (add-hook 'elpy-mode-hook 'flycheck-mode))
```

Elpy's key bindings for flymake won't work, but you can use Flycheck's key bindings then.

tomnor:
OK, thanks, then I know ways around it. I guess I can fiddle with the
key bindings as well.

Will be interesting to see how it turns out later.

pbgc:
@jorgenschaefer 
As I recently started to use 26.1RC I changed to flycheck (I know the keybindings because I use it in several other modes) using your instructions above. The problem is that I get a lot less syntax checks ... it seems that none from flake8.
Am I missing something ?
Thanks in advance!

jorgenschaefer:
I do not know how you configured flycheck and what it uses to check errors. Maybe it found pylint and that's reporting stuff? You could look at the flycheck documentation?

pbgc:
@jorgenschaefer 
Sorry, I assumed that elpy had some self configuration for flycheck (like it seemd to have for flymake).
I will look at the flycheck documentation and and try to configure flycheck.
Thanks for your help!!

[...]

PR #1362, opened 2018-05-01
elpy.el: prepare for flymake in emacs >= 26

tomnor:
flymake-err-info is removed without alias in flymake emacs >= 26.

Related discussions:
https://github.com/jorgenschaefer/elpy/issues/1272
https://github.com/jorgenschaefer/elpy/issues/1357
http://lists.gnu.org/archive/html/bug-gnu-emacs/2018-04/msg00985.html

# PR Summary

Hello Elpy hackers, lets prepare for emacs 26 :)

# PR checklist

## For new features only:

Wrap current code in the first of two cond clauses, add new check for emacs >= 26 in the second clause.

This is a possible refinement of https://github.com/jorgenschaefer/elpy/pull/1360 which I think might fail for emacs < 26 by returning nil. I didn't add or fix any tests for now because I don't feel ready for it, the test frame work is new to me. But tests will need to be updated too I think, the test `elpy-flymake-show-error` is calling `flymake-ler-make-ler` which is gone as well in emacs >= 26.

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -3695,14 +3695,18 @@ description."
 
 (defun elpy-flymake-error-at-point ()
   "Return the flymake error at point, or nil if there is none."
-  (when (boundp 'flymake-err-info)
-    (let* ((lineno (line-number-at-pos))
-           (err-info (car (flymake-find-err-info flymake-err-info
-                                                 lineno))))
-      (when err-info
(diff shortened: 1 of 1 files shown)


jorgenschaefer:
Thank you very much for your contribution!

Issue #1137, opened 2017-05-24
Annoying "version mismatch" popups


offby1:
I recently upgraded the elisp component of elpy from `http://jorgenschaefer.github.io/packages/` and got version 1.15.1.  The most current version of the python component is 1.15.0.  And thus I occasionally see a popup that says
```
Elpy Version Mismatch                                                                                                                                                                                                                                                                                                                                

You are not using the same version of Elpy in Emacs Lisp compared to
Python. This can cause random problems. Please do make sure to use
compatible versions.

Elpy Emacs Lisp version: 1.15.1
Elpy Python version....: 1.15.0
```

This is annoying, since the only way I can think of that will make it shaddap is to downgrade the elisp.

jorgenschaefer:
The Elpy Emacs package comes with the Python code, you should not have to install the elpy package in the virtualenv at all! :-)

jerryasher:
I'm running into the same problem as @offby1 and I don't understand

> The Elpy Emacs package comes with the Python code, you should not have to install the elpy package in the virtualenv at all! :-)

On Win 7 64, using cygwin 32 and cygwin python2.7 and with the python elpy package NOT installed, when `elpy-config` reports

```
Elpy Configuration

Virtualenv........: None
RPC Python........: 2.7.13 (c:/cygwin/bin/python2.7.exe)
Interactive Python: ipython (not found)
Emacs.............: 25.1.1
Elpy..............: Not found (Python), 1.15.1 (Emacs Lisp)
Jedi..............: 0.10.2
Rope..............: Not found (0.10.5 available)
Importmagic.......: 0.1.7
Autopep8..........: 1.3.1
Yapf..............: 0.16.2
Syntax checker....: Not found (flake8)

You have not activated a virtual env. While Elpy supports this, it is
often a good idea to work inside a virtual env. You can use M-x
pyvenv-activate or M-x pyvenv-workon to activate a virtual env.

The Python interpreter could not find the elpy module. Make sure the
module is installed.

[run] python2.7 -m pip install --user elpy

The configured syntax checker could not be found. Elpy uses this
program to provide syntax checks of your programs, so you might want
to install one. Elpy by default uses flake8.

[run] python2.7 -m pip install --user flake8
```

Since `elpy-config` is telling me to install elpy, I guess I will...?

```
$ pip2 install elpy
Collecting elpy
  Using cached elpy-1.15.0-py2.py3-none-any.whl
[...]

PR #1376, opened 2018-05-24
Add some details on the version mismatch error message

galaunay:
# PR Summary
Just add some details on the version mismatch error message (see #1137).
This should makes version mismatch easier to fix for users.


jorgenschaefer:
Very nice! :-)

Issue #1377, opened 2018-05-25
Is `black` autoformatter included?


kootenpv:

galaunay:
At the moment, only `yapf` and `autopep8` are supported,
but it should no be too complicated to support black as well.
What do you think black will do better than yapf ?

note: there is this package also [blacken](https://github.com/proofit404/blacken).

jorgenschaefer:
If someone wants to work on adding support for that, works for me, but I don't see it as a particularly high priority. (Always remember, too, that code we add to the repository also needs to be supported in the future=

gopar:
Black is now included

PR #1378, opened 2018-05-25
Add black autoformatter

galaunay:
# PR Summary
Add [Black](https://github.com/ambv/black) auto formater support to elpy.
Follow issue #1377.
Try it with `M-x elpy-black-fix-code`.

@jorgenschaefer It was a pretty straightforward implementation, mainly copy and paste from yapf implementation.
Not much work involved, so we can perfectly discard this PR if you think it involve adding too much complexity (and support work in the futur).

# PR checklist

## For new features only:
- [x] An entry in NEWS.rst has been added

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,9 @@ install:
     fi
   - pip install autopep8
   - pip install yapf
-  - pip install black
+  - if [[ "$TRAVIS_PYTHON_VERSION" > 3.6 || "$TRAVIS_PYTHON_VERSION" == 3.6 ]] ; then
+      pip install black ;
+    fi
   - pip install coveralls
 script:
   - nosetests
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,7 @@ install:
     fi
   - pip install autopep8
   - pip install yapf
+  - pip install black
   - pip install coveralls
 script:
   - nosetests
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -1,3 +1,8 @@
+New in Elpy development
+=======================
+
+- Elpy now supports Black to format your code.
+
 New in Elpy 1.21.0
 ==================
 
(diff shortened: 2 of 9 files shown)


.travis.yml
```diff
@@ -33,6 +33,7 @@ install:
     fi
   - pip install autopep8
   - pip install yapf
+  - pip install black
```
gopar: Should be a conditional on only 3.6 and up. 
For now we can just make it a conditional to only install 3.6 (After 3.7 is out it'll need to be updated)

gopar:
LGTM

Looks like the CI is only failing because black is not supported on versions less than 3.6.
Other than that should be good to go.

Also, sorry @galaunay for not being as active anymore. I took a break to prioritize my health

NEWS.rst
```diff
@@ -1,6 +1,11 @@
 New in Elpy 1.20.0
 ==================
 
+- Elpy now supports Black to format your code.
```
jorgenschaefer: There is no need to edit NEWS.rst, I do that usually by going through the commit messages when doing a release. That also reduces the confusion with the version numbers here. (Not that I mind people doing this … I'm open for ideas on how to format this? Maybe use "New in Elpy development" above?)

galaunay: "New in Elpy development" seems fine indeed.
I don't know how painful it is to add new features at each releases.
Keeping track of them as and when PR are merged may ease the process ?

I In fact, I realize that "adding en entry to NEW.rst" is one of the bullet point I added on the PR template (see the end of my first post in this PR).
We may want to get rid of that if you want to add entries at each release.

gopar: We can have a separate PR that's only to update NEWS.rst as we go and merge once elpy has a new version. Then repeat the process again? Seems less burdensome that going through commit messages?

PR #1387, opened 2018-06-05
Remove redundant flake8 checks for emacs >= 26.1

### LINKED ISSUE #1386: Unable to silence flymake
As is customary in my ipython config file I have a `get_config()` reference to a non imported symbol. This is not an error and I want to silence the linter using `# noqa`.

Now, this isn't working but I think it's because two different checkers are being run. `flake8` is indeed honouring `# noqa` but the second checker is not.

My supposition about the second checker is based on:

1. Without `#noqa` the error is reported twice, one of the reports include the flake8 code the other just the message.
2.  With `#noqa` the error is reported once without the flake8 code.
3. Style/convention warnings are completely suppresed by `#noqa`.

galaunay:
# PR Summary
Fix #1386.

For emacs >= 26.1, python.el supports flymake.

With both elpy and python.el performing the syntax check, we end up with redundant syntax check errors.

This commit removes a part of elpy's flymake module (only for emacs >= 26.1) to avoid
performing the syntax check twice.

# PR checklist

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -3687,12 +3687,15 @@ display the current class and method instead."
     (`global-init
      (require 'flymake)
      (elpy-modules-remove-modeline-lighter 'flymake-mode)
-     ;; Flymake support using flake8, including warning faces.
-     (setq python-check-command elpy-syntax-check-command)
+     ;; Add our initializer function.
+     ;; For emacs > 26.1, python.el natively supports flymake,
+     ;; so we just tell python.el to use flake8
+     (if (version<= "26.1" emacs-version)
+         (setq python-flymake-command '("flake8" "-"))
+       (setq python-check-command elpy-syntax-check-command)
+       (add-to-list 'flymake-allowed-file-name-masks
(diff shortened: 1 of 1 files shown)


jorgenschaefer:
You sometimes check for <= 26.1 and sometimes for < 26.1 – if that's not intentional, you could merge some of the checks to big when checks for easier reading. Otherwise, looks good to me!

galaunay:
I tried to tidy up a bit and I added some comments for the distant futur when Elpy will stop supporting Emacs < 26.1.

Note to myself: need to check that it still works properly in Emacs < 26.1 before merging.

galaunay:
It works properly for emacs 25.3 as well, merging.

jorgenschaefer:
(Not specific to this PR, but just taking the opportunity here) – You are absolutely awesome. Thank you again for all your work. :-)

galaunay:
Thanks, 
hope it helps :).

Issue #1388, opened 2018-06-07
Signatures not shown in minibuffer


memeplex:
I'm using emacs 26.1 with elpy 1.21.0, the eldoc module is enabled, jedi in installed, autocompletion works.

But I get no signature info at the modeline.

memeplex:
Also, eldoc-mode is enabled in the buffer.

memeplex:
I think the problem is here

```
    (if flymake-error
        flymake-error
```

flymake-error is "", not nil, this might be related to the reworking of flymake in 26.1.

memeplex:
Yeah, it is that, as a workaround I changed that to:

```elisp
    (if (not (string= flymake-error ""))
        flymake-error
```

kootenpv:
I also had this issue. Got the latest version and eldoc stopped showing. This is my elpy config in case it might help:

```
Elpy Configuration

Virtualenv........: mypython--KaCTqkX (/home/pascal/.virtualenvs/mypython--KaCTqkX)
RPC Python........: 3.6.5 (/home/pascal/.virtualenvs/mypython--KaCTqkX/bin/python)
Interactive Python: /home/pascal/python/bin/ipython (/home/pascal/python/bin/ipython)
Emacs.............: 27.0.50
Elpy..............: 1.21.0
Jedi..............: 0.9.0 (0.12.0 available)
Rope..............: Not found (0.10.7 available)
Autopep8..........: 1.3.5
Yapf..............: Not found (0.22.0 available)
Black.............: 18.5b1 (18.6b1 available)
Syntax checker....: pylint (/home/pascal/.virtualenvs/mypython--KaCTqkX/bin/pylint)

```

kootenpv:
@memeplex Awesome that you found the work around, I can confirm, this seemingly solves the issue for me, Eldoc is back.

memeplex:
Check also https://github.com/jorgenschaefer/elpy/issues/1389 because it improves the eldoc experience a lot. Maybe it's worth adding it in the same PR,

memeplex:
Since `elpy-flymake-error-at-point` is called at other places maybe this is a safer and less invasive workaround until the bug is properly fixed:

```
(advice-add 'elpy-flymake-error-at-point
	    :filter-return (lambda (x) (if (string= x "") nil x)))
```

galaunay:
I found this issue as well while trying to fix #1389.
This should be fixed soon by #1392.

memeplex:
[...]

PR #1391, opened 2018-06-07
Fix eldoc

kootenpv:
It is not showing in this version due to most likely an error.

Solution as proposed by #1388


galaunay:
flymake can return `nil` for emacs < 26, hence the errors travis is raising.

I found this bug as well before seeing your PR, this should be fixed soon by #1392.

galaunay:
I am closing this PR following merging of #1392.
Please re-open if the issue is not properly fixed.

Issue #1389, opened 2018-06-07
Annoying interference between company and eldoc


memeplex:
I reported this at company's tracker https://github.com/company-mode/company-mode/issues/796.

I don't know what do you think, but disabling company's meta frontend if elpy's eldoc module is enables could be a reasonable option.

memeplex:
Also see my additions to `eldoc-message-commands` described in https://github.com/company-mode/company-mode/issues/797.

This improves the interaction between eldoc and company **a lot**!

I also added `(eldoc-add-command-completions "python-indent-dedent-line-backspace")`.

memeplex:
@galaunay I'm failing to see anything related to eldoc-message-commands in https://github.com/jorgenschaefer/elpy/pull/1393, that's the only reason I'm not suggesting to close this issue.

PR #1393, opened 2018-06-09
Prevent interferences between eldoc and company

galaunay:
# PR Summary
Follow issue #1389.

When eldoc and company modules are enable, and company activated, the two modules compete for the minibuffer message.
This create flickering and changing minibuffer message.

This PR deactivate company meta frontend, and indicate to eldoc to reprint message after company functions to avoid those issues.

# PR checklist

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -3633,6 +3633,10 @@ or unless NAME is no callable instance."
      (require 'eldoc)
      (setq eldoc-minor-mode-string nil))
     (`buffer-init
+     ;; avoid interferences between eldoc and company meta frontend
+     (when (member 'elpy-module-company elpy-modules)
+       (set (make-local-variable 'company-frontends)
+            (delq 'company-echo-metadata-frontend company-frontends)))
      (set (make-local-variable 'eldoc-documentation-function)
           'elpy-eldoc-documentation)
      (eldoc-mode 1))


elpy.el
```diff
@@ -3633,6 +3633,11 @@ or unless NAME is no callable instance."
      (require 'eldoc)
      (setq eldoc-minor-mode-string nil))
     (`buffer-init
+     ;; avoid interferences between eldoc and company meta frontend
+     (when (member 'elpy-module-eldoc elpy-modules)
+       (set (make-local-variable 'company-frontends)
+            (delq 'company-echo-metadata-frontend company-frontends))
+       (eldoc-add-command-completions "company-"))
```
memeplex: I'm also removing python-indent-dedent-line-backspace here since I found it to play bad with company.

BTW, I submitted an issue to debbugs asking for eldoc to change its whitelist approach for one that just checks if the echo area has no message before writing. This heuristic doesn't depend on a list of commands. Although I'm probably wrong somehow.

elpy.el
```diff
@@ -3633,6 +3633,12 @@ or unless NAME is no callable instance."
      (require 'eldoc)
      (setq eldoc-minor-mode-string nil))
     (`buffer-init
+     ;; avoid interferences between eldoc and company meta frontend
+     (when (member 'elpy-module-eldoc elpy-modules)
```
jorgenschaefer: Shouldn't this always be the ase here?

galaunay: Yes, you are right, I meant `elpy-module-company`.
let me change that.

galaunay:
WIP:
Deactivating company meta frontend is a solution to avoid elpy and company fighting to display their infos with eldoc.
But I would bet better to find a way to make them cooperate and having both infos at the right moment.

memeplex:
I understand your concern @galaunay but then I would say that this should be splitted into two changes:

1. Avoid message **switching**: deactivate company meta frontend by default. This change is slightly more controversial, maybe it's worth a customization option or, at least, a remark in the documentation showing how to deactivate the frontend from a python-mode-hook.

2. Avoid message **flickering**: whitelist the company commands (and some others). This makes all the difference in terms of flickering. It's not a controversial change IMO and, in fact, it should be implemented at company level (I've already requested that). But, for the time being, adding `(eldoc-add-command-completions "company-" "python-indent-dedent-line-backspace")` significatively improves the eldoc experience.

So, my suggestion is to go on with 2 and keep 1 as WIP.

galaunay:
After taking a closer look, It is difficult to make eldoc and company-metadata cooperate and display the right info at the right time.

Anyway, as company metadata infos are relatively useless in python (just displaying info on the type: class, function, variable), I think deactivating it in profit of eldoc is the right thing to do.

I am merging this.

Issue #240, opened 2014-06-01
Module: Modern code folding


jorgenschaefer:
Elpy should support code folding like current IDEs provide. That is, simple code folding commands for sensible blocks of Python code (indentation blocks?), with markers and interactable icons in the fringe.
- `hs-toggle-hiding` should do most of that for us
- Needs a good key shortcut
  - C-TAB maybe. Doesn't work in console. backtab there?
  - Or just TAB at the beginning of a line? Shouldn't be ambiguous.
  - Atom and Sublime Text use C-M-[ / C-M-], with shift for "everything". And people say Emacs has crazy key bindings.
    - With region, folds that region
    - Anywhere in a block of code, folds that block
- Prefix argument should hide leaf nodes (or lines with "def" in them)
- Visualization in the fringe is provided by http://www.emacswiki.org/emacs/hideshowvis.el but that's very limited, won't even auto-update

cpitclaudel:
Indeed, folding docstrings would be very nice :)

humitos:
Right now I'm using origami (https://github.com/gregsexton/origami.el) with these settings

```
(require 'origami)

(define-key origami-mode-map (kbd "C-c f a") 'origami-toggle-all-nodes)
;; http://stackoverflow.com/questions/916797/emacs-global-set-key-to-c-tab
(define-key origami-mode-map (kbd "<C-tab>") 'origami-recursively-toggle-node)

(global-origami-mode 1)
```

Yeah, it's not the best and I would prefer something better than that but it's something that could be considered in the research.

aparkerlue:
I find that the following is an acceptable stopgap:

```emacs-lisp
(add-hook 'python-mode-hook 'hs-minor-mode)
(with-eval-after-load "python"
  (define-key python-mode-map (kbd "C-c TAB") 'hs-toggle-hiding)
  )
```

clawdelle:
@aparkerlue thx, that is coll,enough for me

PR #1398, opened 2018-06-19
WIP: Getting folding right in Elpy

galaunay:
# PR Summary
Follow (old) feature request #240.

This is an attempt to get nice folding functionalities in Elpy.
It is currently based on `hs-minor-mode` and heavily inspired from [hideshowvis](https://www.emacswiki.org/emacs/hideshowvis.el).

Any idea is welcome.

# Todo list
## Nice folding
- [x] Add an overlay indicating the number of folded lines.
- [x] Add a fringe indicator for folded code.
- [x] Make fringe indicators clickable 
- [x] Make folding indicators clickable
- [x] Add fringe indicators to mark the beginning of foldable regions.
- [ ] Add fringe indicator for dosctrings ?
## Folding docstrings
- [x] Add `elpy-folding-toggle-docstrings` to hide all the docstrings in the buffer.
- [x] Add `elpy-folding-hide-docstring-at-point` functions to hide the docstring at point.
- [x] Find a way to show the first docstring line when folding docstring.
## Folding comments
- [x] Add a function to hide all comments
## DWIM
- [x] Add a function to toggle folding dor the thing (block, docstring, region or comment) at point
## Selective folding
- [x] Add `elpy-folding-hide-leafs` to hide blocks not containing other blocks.
## Keybindings
- [x] Add `elpy-folding-fold-at-point` to fold the active region, the docstring,  or the block at point.
- [x] Bind `elpy-folding-toggle-at-point` (to `C-c @ p`).
- [x] Bind `elpy-folding-toggle-all-docstrings` (to `C-c @ b`).
- [x] Bind `elpy-folding-hide-leafs` (to `C-c @ f`).
## Other modules incompatibilities
- [x] Put flymake fringe indicators on the right fringe when folding module is activated to avoid fringe indicator overlap.
## Others
- [x] Clean and reorganize properly
- [x] Write tests
- [x] Update the documentation

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -90,6 +90,8 @@ can be inidividually enabled or disabled."
                      elpy-module-eldoc)
               (const :tag "Highlight syntax errors (Flymake)"
                      elpy-module-flymake)
+              (const :tag "Code folding"
+                     elpy-module-folding)
               (const :tag "Show the virtualenv in the mode line (pyvenv)"
                      elpy-module-pyvenv)
               (const :tag "Display indentation markers (highlight-indentation)"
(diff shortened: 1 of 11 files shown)


galaunay:
@jorgenschaefer, before I begin to add tests and documentation, do you have any feedback on this ?

Notably, I am not sure about the keybindings.

Also, It is inspired from [here](https://www.emacswiki.org/emacs/HideShow) for the leafs hiding and [hideshowvis](https://github.com/sheijk/hideshowvis) (GPL v2) for the fringe and folding overlays. Not sure if we need to add that as source somewhere.

jorgenschaefer:
This looks really nice. The only general feedback I have is "are you sure you want to maintain this in the future" – any such larger code base will need maintenance to keep up to date with dependent packages. But other than that, nice.

The key bindings are pretty standard for Emacs. I never liked C-c @, because @ is "right alt + q" on German keyboards, but that's a general problem. Being standard, I think this is fine.

Thank you again for all the great work you do!

galaunay:
This is a good advice, I will try to keep it simple.
Notably, I am not sure that mouse integration is really necessary.

I'll stick to the standard key bindings then (not really convenient on the french keyboard as well, but I'll rebind that).

peterewills:
Thanks for building this! 🙏 I've wanted code folding in elpy for a while, and now I'm running your branch and it works well. Would be nice to have this in master!

galaunay:
Thanks for the feedback.
I kept this PR on the side because of lack of time for proper testing.
Apart of that, I think it is kind of ready to be merged.

So If you are saying it's working, I could give it a quick test and merge it.
Any additional features you would find useful ?

peterewills:
I've actually been having some weird issues with `hs-hide-level`. These don't appear when I just use `hs-minor-mode` on top of the MELPA version of elpy. 

Essentially, when I run `hs-hide-level 2`, it only hides every other block at level 2. This only appears on my (larger & more complicated) work files, which I can't share here, so I'll have to take some time to try and build a minimal example. I made a very simple .py file with a number of levels, and that seems to work fine, so it's something in the way that things get parsed on larger files with docstrings, comments, etc.

I'll pull down this branch at the latest commit and make sure I'm still seeing the bug, and then try to get a minimal example up here.

As for features: The only thing that I want is a function that does "hide all blocks that are either a leaf or at level n", which is like `elpy-folding-hide-leafs` but won't expand any blocks beyond level n. That's kind of idiosyncratic to my preferences, though, so if you think I should just build that for myself that's totally reasonable.

elpy.el
```diff
@@ -3058,6 +3073,380 @@ display the current class and method instead."
...
+;;     taken from https://www.emacswiki.org/emacs/HideShow
+(defun elpy-folding--hide-leafs (beg end)
+  "Hide blocks that do not contain others blocks in region (BEG END)."
+  (save-excursion
+    (goto-char beg)
+    ;; Find the current block beginning and end
+    (when (hs-find-block-beginning)
+      (setq beg (1+ (point)))
+      (funcall hs-forward-sexp-func 1)
+      (setq end (1- (point))))
```
peterewills: I think this needs to move the pointer back to the beginning of the block here.
```suggestion
      (setq end (1- (point)))
      (goto-char beg))
```

galaunay:
@peterewills I finally managed to finish and merge this PR.

The folding module is deactivated by default, you will have to activate it trough customize or by adding it to `elpy-modules`.
Hope you will find it useful.

azzamsa:
instead of having independent module for folding, why don't we just tell user to use external package such us origami https://github.com/gregsexton/origami.el ?

PR #1404, opened 2018-06-23
Avoid eldoc message flickering

galaunay:
# PR Summary
Follow discussion in #1393.

Remove annoying flickering when using eldoc and company modules jointly.


elpy.el
```diff
@@ -3633,6 +3633,10 @@ or unless NAME is no callable instance."
      (require 'eldoc)
      (setq eldoc-minor-mode-string nil))
     (`buffer-init
+     ;; avoid eldoc message flickering
+     (when (member 'elpy-module-company elpy-modules)
```
jorgenschaefer: This can be `memq` … I'm not sure I really like this condition. Is it harmful to add the following line even if elpy-module-company is disabled?

galaunay: I don't think so.
But in that case maybe we can restrain the number of company function added to `eldoc-message-commands` to avoid side-effects.

galaunay:
@memeplex do you think all the `company-` functions need to be put in `eldoc-message-commands`. 
It would be better to just add the necessary functions.

memeplex:
No company command writes to the echo area, company only writes to the echo area when triggered by a timer, so the right thing to do is to whitelist company commands as "don't write". See further discussion here https://github.com/company-mode/company-mode/issues/797.

galaunay:
As it is currently being solved upstream, I will wait before merging anything.
Thank you for the info @memeplex.

memeplex:
I'm not really sure what will happen upstream, they have more ambitious
plans as it seems. I would say that you add it for now, it does no harm at
all.

Issue #1412, opened 2018-07-13
Change `C-c C-y C-g` key binding


aparkerlue:
`(elpy-shell-send-group-and-step)` is currently bound to `C-c C-y C-g`. This is an unfortunate choice, because `C-g` is the standard Emacs binding for `(keyboard-quit)`, and the natural keystroke to abort n the middle of a `C-c C-y` execution statement is `C-g`.

I suggest finding a different key (other than `g`) for executing "groups" of Python statements. I do not have a better alternative right now.

galaunay:
Thanks for the feedback. I guess we could use "o" as a replacement (for "grOup" ?):
```elisp
    (define-key map (kbd "C-c C-y o") 'elpy-shell-send-group)
    (define-key map (kbd "C-c C-y O") 'elpy-shell-send-group-and-go)
    (define-key map (kbd "C-c C-y C-o") 'elpy-shell-send-group-and-step)
    (define-key map (kbd "C-c C-y C-S-O") 'elpy-shell-send-group-and-step-and-go)
```

Do you think it would be a good solution ?

galaunay:
I made a PR for this (#1415).

I went for `o` but I don't have a strong opinion on this either.
It could have been 'b' for block, but its already taken for sending the buffer...

galaunay:
It's merged.

Thank you for bringing that up.
I guess I can close this now.

PR #1415, opened 2018-07-19
Replace keybinding to avoid using 'C-g'

galaunay:
# PR Summary
Following issue #1412.

Change the keybinding of function sending the current group to the shell.
The previous keybindings would use `C-g` which is also the keybinding used to abort a set of key press.

# PR checklist


gopar:
LGTM

Issue #1416, opened 2018-07-19
ElDoc module ignores eldoc-echo-area-use-multiline-p


bpmcl:
# Summary
When using elpy-mode with the ElDoc module enabled, I've noticed that long function and definition signatures can cause the minibuffer echo area to be resized.

After investigation, I found that there is a variable `eldoc-echo-area-use-multiline-p` that restricts this behaviour. However, even after setting it to `nil`, the issue was still occurring. The documentation for this variable states:

> Note that this variable has no effect, unless ‘eldoc-documentation-function’ handles it explicitly.

Is it possible to add handling for this to the elpy ElDoc module?

# Steps to reproduce
Highlight a function/definition that has long documentation with the ElDoc module enabled. Set the `eldoc-echo-area-use-multiline-p` variable to `nil` and re-attempt, but no change will happen.

# My configuration
## OS
macOS 10.13.6
iTerm2 3.1.7
GNU Emacs 25.3.1 (x86_64-apple-darwin17.0.0)
elpy 20180707.1413

## Result of `(elpy-config)`
```
Virtualenv........: None
RPC Python........: 3.5.4 (/Users/i055459/.pyenv/shims/python)
Interactive Python: python (/Users/i055459/.pyenv/shims/python)
Emacs.............: 25.3.1
Elpy..............: 1.22.0
Jedi..............: 0.12.1
Rope..............: 0.10.7
Autopep8..........: 1.3.5
Yapf..............: 0.22.0
Black.............: Not found (18.6b4 available)
Syntax checker....: flake8 (/Users/i055459/.pyenv/shims/flake8)
```

## Elpy configuration in my init.el
```
 '(eldoc-echo-area-use-multiline-p nil)
 '(eldoc-idle-delay 0.5)
 '(eldoc-minor-mode-string nil)
 '(elpy-eldoc-show-current-function t)
 '(elpy-modules
   (quote
    (elpy-module-company elpy-module-eldoc elpy-module-flymake elpy-module-pyvenv elpy-module-highlight-indentation elpy-module-yasnippet elpy-module-django elpy-module-sane-defaults)))
```

jorgenschaefer:
Thank you for the report! I'm unsure why Eldoc does not check that variable by itself, but if it does not, Elpy should probably respect that …

bpmcl:
[...]

PR #1417, opened 2018-07-20
Male eldoc message shorter and truncate them if too long

galaunay:
# PR Summary
Follow #1416: 

- Elpy eldoc messages are now truncated to fit in the minibuffer if `eldoc-echo-area-use-multiline-p` is `nil`.

- Eldoc messages are now shorter, removing the useless `param ` strings returned by the jedi backend.

# PR checklist

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -3670,6 +3670,7 @@ display the current class and method instead."
            ((stringp calltip)
             calltip)
            (t
+            (message "OK1")
             (let ((name (cdr (assq 'name calltip)))
                   (index (cdr (assq 'index calltip)))
                   ;; Strip 'param' added by jedi at the beggining of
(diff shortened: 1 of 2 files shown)


bpmcl:
While this mostly works, for some reason the ElDoc output still wraps to another line in my setup, so it looks like this:

```python
ClassName(var1=None, var2=None, ..., va/
r
```

I'm unsure if the `minibuffer-window` size is returning incorrectly in my specific setup, but changing the width expression to the following fixed it for me:

```elisp
(- (window-width (minibuffer-window)) 1)
```

@galaunay Did it display correctly in your setup without this change?

galaunay:
I finally found an eldoc function to format properly the eldoc messages.
It should work properly now.

(and @bpmcl , you were right, it uses `(1- (window-width (minibuffer-window)))`).

Issue #1424, opened 2018-07-31
Feature request: Only switch to shell when there's output


Yevgnen:
Hi, thanks for the awesome package. I wonder if it is possible to switch to shell only when there's output after evaluation? It seems that currently `elpy-shell-display-buffer-after-send` can only `t` or `nil` right now.

Or can one display the output in minibuffer? The [lpy](https://github.com/abo-abo/lpy) package does this and will display `(ok)` in the minibuffer when there's no output.

Thanks.

galaunay:
Hi, 

After having a quick look, 
Your first proposition is not a trivial thing to do.

But you can have the output to be displayed in the minibuffer using the `elpy-shell-echo-output` option. It will echo `no output was produced` if there is not output to display.

Hope it helps.

Yevgnen:
@galaunay Thanks for the information about `elpy-shell-echo-output`!

Yevgnen:
@galaunay The `elpy-shell-echo-output` variable seems also take control in a python shell. When I hit enter in python shell, the output also echoes!

galaunay:
I just tested it, and you are right.
Thanks for noticing, this is a proper bug.

I made a PR (#1427) to fix that.
This is not a big PR and it should be merged soon.

Yevgnen:
@galaunay Thanks for the quick following up.

rgemulla:
Just saw this and the PR now. 

The fix in #1427 causes issues when multiple statements are queued. E.g., when evaluating the two calls of ```f``` in the piece of code below (with the second one before the first one finishes), only the output of the first call to ```f``` is echoed. I.e., the output of the second call to ```f``` is incorrectly omitted.

```python
import time

def f(s):
    time.sleep(3)
    return s

f("call 1")

f("call 2")
```

In general, I feel that echoing output can also be useful if a command is entered directly into the shell buffer. For example, it may be useful when the command execution takes a while and the shell buffer is not visible when it ultimately completes. 

@Yevgnen: When ```elpy-shell-echo-output``` is set to ```when-shell-not-visible``` (the default), no echoing takes place as long as the shell is visible. Would this solve your issue as well?

galaunay:
I am not using this feature a lot, so I may have incorrectly understood its use and/or the code.
@rgemulla Feel free to revert this if you think it is the best solution.

rgemulla:
Reverted. 

[...]

PR #1427, opened 2018-08-01
Fix echoing output in the minibuffer

galaunay:
# PR Summary
Follow discussion in #1424.

When `elpy-shell-echo-output` is `t`, the output is always echoed in the minibuffer, even when inputs are directly entered into the shell buffer.

This is due to the variable `elpy-shell--capture-output`, that is set to `t` by commands needing to echo the output, but never set back to `nil`.
Consequently, the output filter grabs and echoes all outputs.

This is fixed here by reseting `elpy-shell--capture-output` to `nil` after having echoed an output.

WIP: need to be tested properly.

# PR checklist

### DIFF
--- a/elpy-shell.el
+++ b/elpy-shell.el
@@ -464,7 +464,9 @@ complete). Otherwise, does nothing."
         (if (string-empty-p output)
             (message "No output was produced.")
           (message "%s" (replace-regexp-in-string "\n\\'" "" output))))
-      (setq-local elpy-shell--captured-output nil)))
+      (setq-local elpy-shell--captured-output nil)
+      ;; Stop capturing the outputs then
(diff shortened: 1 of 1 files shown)


jorgenschaefer:
I just stumbled over `caputre` vs. `captured` – I generally dislike single-character differences (past/present, singular/plural) that are meaningful, and prefer more verbose name differences, but that's not part of this PR, so this one looks good to me :-)

galaunay:
I did as well while hunting for the bug...
If it does look good to you, I guess we can safely merge this without @rgemulla feedback.

rgemulla:
I suggest to roll back this PR (see my comment in #1424). We may either keep things as they were before or look for another solution.

Issue #1426, opened 2018-08-01
xref-find-definitions is really slow with prefix arg


raxod502:
# Summary

When I use `M-.` in Elpy, it's very fast. However, `C-u M-.` takes several seconds to pop up the list of options. There is no similar delay when I use `C-u M-.` in Elisp.

This creates a poor user experience.

# Steps to reproduce

* Run `emacs -Q`.
* Load `elpy` revision 0086773f643cedfe52e847f7d542e3c6636c730a.
* Run `(setq elpy-rpc-python-command "python3")`
* Run `M-x elpy-enable`.
* Open a decent-sized Python file, e.g. https://github.com/raxod502/diary-manager/blob/ab7fd57de75354a9208470b45ab7bb17a26b95f4/diary which is what I used to test.
* Find a symbol usage, e.g. invocation of `main` at https://github.com/raxod502/diary-manager/blob/ab7fd57de75354a9208470b45ab7bb17a26b95f4/diary#L336 which is what I used to test.
* Press `C-u M-.`.
* Enjoy several seconds of waiting before the completions menu is populated.

# My configuration
## OS

macOS 10.11.6

## Result of `(elpy-config)`
```
Virtualenv........: None
RPC Python........: 3.7.0 (/usr/local/bin/python3)
Interactive Python: python (/usr/local/bin/python)
Emacs.............: 26.1
Elpy..............: 1.23.0
Jedi..............: 0.12.1
Rope..............: 0.10.7
Autopep8..........: 1.3.5
Yapf..............: 0.22.0
Black.............: 18.6b4
Syntax checker....: flake8 (/usr/local/bin/flake8)
```

## Elpy configuration in my init.el

Not applicable, since I'm running from `emacs -Q`.

galaunay:
Thanks for the report.

The delay is due to the fact that Elpy checks each candidate definition to avoid proposing several candidates leading to the same definition. While this is convenient (no redondant candidates), it can indeed be quite slow for big files.

I made PR #1429 to remove this feature (you are right, this is quite unusable as it is).

galaunay:
As an extra safety, could you have a quick go at #1429 ?
Just to check to its still behaving as expected for you.

galaunay:
@raxod502 thanks for the feedback.
Please let us know if something weird/wrong happen following the merging.

PR #1429, opened 2018-08-05
Make xref faster

galaunay:
# PR Summary
Follow #1426.

In xref, do not check for candidate assignments anymore.

This feature was present to prevents redondant candidates,
but is too slow to be decently usable.

# PR checklist

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -3153,17 +3153,11 @@ Points to file FILE, at position POS."
     "Return identifier at point.
 
 Is a string, formatted as \"LINE_NUMBER: VARIABLE_NAME\".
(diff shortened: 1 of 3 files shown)


jorgenschaefer:
lgtm. If duplicates are an issue, maybe do the deduplication in Python, or use a hash table instead of a list for uniqueness checks?

galaunay:
What I mean here by 'duplicates' are symbols with the same name, at different lines and leading to the same definition.
In the 'goto_definition' context, having several of them is quite redondant.

Unfortunately, it seems difficult to remove them without having to properly check their definition (which is way too long...).

Issue #1274, opened 2018-01-23
Neither easy_install nor pip found


Hugozys:
When I type M-x elpy-config in emacs froze a seconds, then showed the following messages:
**Neither easy_install nor pip found**

Then I typed 
sudo apt-get install python3-pip
sudo apt-get install python3-setuptools

The prompt showed the following messages:
Reading package lists... Done
Building dependency tree
Reading state information... Done
***python3-pip/python3-setuptools*** is already the newest version (8.1.1-2ubuntu0.4).
The following packages were automatically installed and are no longer required:
  accountsservice gir1.2-glib-2.0
  libaccountsservice0 libgirepository-1.0-1
Use 'sudo apt autoremove' to remove them.
0 upgraded, 0 newly installed, 0 to remove and 3 not upgraded.

So I think I already installed easy_install and pip in my Linux server
What should I do to fix this problem?

Hugozys:
After I install python-pip, it worked. But I don't quite get it. My python version 3.5.2 should I install pip3 only instead of pip?

jorgenschaefer:
Hello and thanks for the report! What's the output of `python -m pip --help` for you?

Hugozys:
Usage:
  /usr/bin/python -m pip <command> [options]

Commands:
  install                     Install packages.
  download                    Download packages.
  uninstall                   Uninstall packages.
  freeze                      Output installed packages in requirements format.
  list                        List installed packages.
  show                        Show information about installed packages.
  search                      Search PyPI for packages.
  wheel                       Build wheels from your requirements.
  hash                        Compute hashes of package archives.
  completion                  A helper command used for command completion
  help                        Show help for commands.

General Options:
  -h, --help                  Show help.
  --isolated                  Run pip in an isolated
                              mode, ignoring
                              environment variables
                              and user
                              configuration.
  -v, --verbose               Give more output.
                              Option is additive,
                              and can be used up to
                              3 times.
  -V, --version               Show version and exit.
  -q, --quiet                 Give less output.
  --log <path>                Path to a verbose
                              appending log.
  --proxy <proxy>             Specify a proxy in the
                              form [user:passwd@]pro
                              xy.server:port.
  --retries <retries>         Maximum number of
                              retries each
                              connection should
                              attempt (default 5
                              times).
  --timeout <sec>             Set the socket timeout
                              (default 15 seconds).
  --exists-action <action>    Default action when a
                              path already exists:
                              (s)witch, (i)gnore,
[...]

PR #1442, opened 2018-08-15
Allow pip-related tests to pass with python3 or /usr/bin/python3

sten0:
# PR Summary
I believe this is closely related to #1274, and further changes may be required, because this patch only affects the self tests.  This patch is needed to prevent the following failing tests when building on a Debian 10 (buster, though technically I'm building for sid) where /usr/bin/python3 is the only Python.
```
2 unexpected results:
   FAILED  elpy-config-should-show-flake8-pip-button-when-no-syntax-checker-available
   FAILED  elpy-insert--pip-button-value-create-should-use-pip
```
Here is how I added preliminary configuration to default to system-installed Python 3 in the Debian (and inherently the future Ubuntu) package:
https://salsa.debian.org/emacsen-team/elpy/blob/master/debian/debian-config.el
https://salsa.debian.org/emacsen-team/elpy/blob/master/debian/elpa-test
# PR checklist

### DIFF
--- a/test/elpy-config-test.el
+++ b/test/elpy-config-test.el
@@ -21,4 +21,4 @@
                                          (elpy/wait-for-output "Only")
                                          (buffer-string))))
                            (should (equal python-check-command "/foo/bar/flake8"))
-                           (should (string-match "python -m pip install.*flake8" output))))))
+                           (should (string-match ".*python.* -m pip install.*flake8" output))))))
--- a/test/elpy-config-test.el
+++ b/test/elpy-config-test.el
@@ -21,4 +21,4 @@
                                          (elpy/wait-for-output "Only")
                                          (buffer-string))))
                            (should (equal python-check-command "/foo/bar/flake8"))
-                           (should (string-match ".*python.* -m pip install.*flake8" output))))))
+                           (should (string-match "python.*\ -m pip install.*flake8" output))))))
--- a/test/elpy-insert--pip-button-value-create-test.el
+++ b/test/elpy-insert--pip-button-value-create-test.el
@@ -17,8 +17,8 @@
               (program &optional infile destination display &rest args)
               0)
              (widget (widget-create 'elpy-insert--pip-button "test-module")))
-      (should (string-match ".*python.* -m pip" (widget-get widget :command)))
-      (should (string-match ".*python.* -m pip" (buffer-string))))))
+      (should (string-match "python.*\ -m pip" (widget-get widget :command)))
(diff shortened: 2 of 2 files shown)


galaunay:
It seems fine for me.

Do you wish to merge that now, or should we wait for potential further changes ?

sten0:
As you prefer :-)  I plan to include this patch and also enable a "works out of the box defaulting to system Python3 interpreter and libs" in the next Debian upload...unfortunately that plan is currently blocked by this yapf + Python 3.7 incompatibility issue https://github.com/google/yapf/issues/586 so it might be a while before I return to this.

AndreaCrotti found issues with our snippets merge so I'm going to prioritise working on that, because the issues are probably my fault ;-)

test/elpy-config-test.el
```diff
@@ -21,4 +21,4 @@
                                          (elpy/wait-for-output "Only")
                                          (buffer-string))))
                            (should (equal python-check-command "/foo/bar/flake8"))
-                           (should (string-match "python -m pip install.*flake8" output))))))
+                           (should (string-match ".*python.* -m pip install.*flake8" output))))))
```
galaunay: Is the `.*` at the beginning really useful ?
`string-match` should return non-nil for a regexp anywhere in the string.

test/elpy-insert--pip-button-value-create-test.el
```diff
@@ -17,8 +17,8 @@
               (program &optional infile destination display &rest args)
               0)
              (widget (widget-create 'elpy-insert--pip-button "test-module")))
-      (should (string-match "python -m pip" (widget-get widget :command)))
-      (should (string-match "python -m pip" (buffer-string))))))
+      (should (string-match ".*python.* -m pip" (widget-get widget :command)))
```
galaunay: Same here ?

test/elpy-config-test.el
sten0: You're right.  I've fixed this in 766ddf9 and additionally tightened the regex to account for the space between the hungry glob and -m.

test/elpy-insert--pip-button-value-create-test.el
sten0: Yes.  See above.

test/elpy-config-test.el
sten0: weird, my comments aren't showing up here... 'hope you got them elsewhere!

galaunay:
Perfect, thanks !

Issue #1431, opened 2018-08-07
Backporting to emacs 24.3 ?


CdeMills:
# Summary

Hello,
would it be possible to backport or make your package compatible with Emacs 24.3 (CentOS, RedHat ?)

# Steps to reproduce
I tried to install your package from melpa (Meta-X install-package elpy) but I get an error telling the package was not available, it seems lowest supported version is 24.4. Using CentOS, upgrading from 24.3 to 24.4 is not an option

# My configuration
CentOS Linux release 7.5.1804 (Core)
Emacs emacs-24.3-20.el7_4.x86_64
## OS
Linux 4.17.11-1.el7.elrepo.x86_64

Regards

Pascal
## Result of `(elpy-config)`
```
# Paste it here
```

## Elpy configuration in my init.el
```
# Paste it here
```

CdeMills:
[elpy-backport-24.3.pat.txt](https://github.com/jorgenschaefer/elpy/files/2269549/elpy-backport-24.3.pat.txt)

Hello,
I produced a patch for the issue:
1) define subr-x if not available
2) disable profiling is make-process is not available

I created a local git repo; made a "tar" image of it; but M-x package-install-file is not working, there is some issue with the filename.

Could you test ?

Regards

Pascal

galaunay:
Hi Pascal, 

I don't think we will add this backport to the master branch of Elpy, 
but I guess we can make a specific branch for this.

Could you provide your tar image, so I can test on my side ?

galaunay:
I couldn't find your tar file in the issue.
Has it been filtered by github ?

CdeMills:
[elpy-git-tar.b64.gz](https://github.com/jorgenschaefer/elpy/files/2277049/elpy-git-tar.b64.gz)

Hello,
so it seems. 

The requested file must be decompressed with gzip, then uudecoded from base64

Regards

Pascal

galaunay:
For Emacs to be able to install the package with `M-x package-install-file`, the tar file should extract in a folder with the name `elpy-20180803.1438` (`20180803.1438` being the Elpy version).

You can check the Emacs doc on that [here](https://www.gnu.org/software/emacs/manual/html_node/elisp/Multi_002dfile-Packages.html#Multi_002dfile-Packages).

I was able to make it work with this one:
[elpy-20180803.1438.tar.zip](https://github.com/jorgenschaefer/elpy/files/2280370/elpy-20180803.1438.tar.zip)
[...]

PR #1444, opened 2018-08-18
Add support for emacs 24.3

galaunay:
# PR Summary
Follow #1431.

Add back support for Emacs 24.3, as proposed by @CdeMills.

@jorgenschaefer I don't like the idea of complexifying the code either, but this is apparently useful for some users.
Let to discuss if we want to merge this into the main branch, or instead to make a backport branch.

# PR checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ python:
 env:
   # - EVM_EMACS=emacs-24.1-travis
   # - EVM_EMACS=emacs-24.2-travis
-  # - EVM_EMACS=emacs-24.3-travis
+  - EVM_EMACS=emacs-24.3-travis
   - EVM_EMACS=emacs-24.4-travis
   - EVM_EMACS=emacs-24.5-travis
   - EVM_EMACS=emacs-25.1-travis
--- a/docs/ide.rst
+++ b/docs/ide.rst
@@ -192,7 +192,7 @@ Interpreter Setup
 Elpy uses the Python interpreter setup from the Emacs ``python`` package. This
 section briefly summarizes some common setups; add the one you need to your
 ``.emacs`` file. Note that the code below (and Elpy in general) require at least
-Emacs 24.4.
+Emacs 24.3.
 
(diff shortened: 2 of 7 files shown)


jorgenschaefer:
I'm still confused why @CdeMills can't just upgrade their Emacs? We do not support Python 2.5, either, for example.

Can't this be solved by some gist with an .el file that users with utterly outdated Emacs can use?

CdeMills:
Hello,

For years, I was under Fedora. This means, every 6 month, the same circus.
New version going out, read the doc, find how to upgrade from n to n+1
without reinstalling, try it, understand why it fails (the reason is
different at each new version), and so on.

CentOS offers a much more stable experience. You have small updates which
never break any compatibility rule. The drawback is, as you mention, that
most of the software can be seen as ... outdated.

I compiled Emacs 26.6 using the Linux From Scratch approach. It runs ...
but I get from time to time segmentation faults. I don't have the time to
dig and understand why. So getting elpy working in conjunction with
Emacs-24.3 is the shortest path, with the added benefit that all peoples
under Redhat / CentOS / Scientific Linux could use it.

Regards

Pascal


Le ven. 31 août 2018 à 10:20, Jorgen Schäfer <notifications@github.com> a
écrit :

jorgenschaefer:
Might I suggest using an older version of Elpy, then?

CdeMills:
Hello,

why not ? What is the latest elpy compatible with 24.3 ? How to install it
from package manager ?

Regards

Pascal

Le ven. 31 août 2018 à 11:19, Jorgen Schäfer <notifications@github.com> a
écrit :

jorgenschaefer:
Emacs' package management system only supports installing the latest version, so you have to use the git repo. You can try the commit `f6a8ac01f0fa191e6f09547fa77d017004fe7b45`. If that does not work, try `1c21d9f8fb8ad8020c60f165ced1ccb753a9ca55`.

galaunay:
This is a duplicate of #1448.
I am closing this.

PR #1458, opened 2018-09-09
fix completion with no python process

gaetano-guerriero:
Completion coming from rpc are intended to merged with completions from python.el.

`python-completion-complete-at-point` returns nil when there is no python shell
process, `elpy-company--add-interpreter-completions-candidates` returns nil in
this case and this shadows completion coming from rpc.

# PR Summary

# PR checklist

## For new features only:


galaunay:
lgtm, thanks for that.

PR #1460, opened 2018-09-10
Default to "." format for django test format

gaetano-guerriero:
This allows to run out-of-the-box in a project where a custom runner is in use
but it is simply a subclass of standard django.test.runner.DiscoverRunner

# PR Summary
In my experience, often the runner in a django project is overridden with a custom runner simply inheriting from `DiscoverRunner`. In this case case it seems an overkill to require the configuration of `elpy-django-test-runner-formats` for every project.
This PR simply let `"."` separator be the default when running django tests

# PR checklist

## For new features only:

### DIFF
--- a/elpy-django.el
+++ b/elpy-django.el
@@ -78,12 +78,11 @@ require arguments in order for it to work."
   :group 'elpy-django)
 (make-variable-buffer-local 'elpy-django-commands-with-req-arg)
 
-(defcustom elpy-django-test-runner-formats '(("django_nose.NoseTestSuiteRunner" . ":")
-                                             ("django.test.runner.DiscoverRunner" . "."))
+(defcustom elpy-django-test-runner-formats '(("django_nose.NoseTestSuiteRunner" . ":"))
   "List of test runners and their format for calling tests.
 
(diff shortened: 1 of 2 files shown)


jorgenschaefer:
I'm generally not too happy with changing prescribed defaults because some people change the defaults on their end. Can't you customize `elpy-django-test-runner-formats` globally to suite your needs?

gaetano-guerriero:
Working on different projects each with its own runner now require to set `elpy-django-test-runner-formats` to `(("myproject.test.TestRunner" . "."))` for every project. 
This is IMHO a bad user experience in using elpy.

jorgenschaefer:
Maybe `elpy-django-test-runner-formats` should allow for regular expressions, then.

@gopar what do you think?

Also, I don't think this custom option should be buffer-local by default, I think making defcustom variables buffer-local like that is bad practice.

gaetano-guerriero:
I'm going to implement the regex solution so this can be closed.

Anyway should (".*" . ".") be the least priority entry in `elpy-django-test-runner-formats` ?

gopar:
Yeah, I think it should be the last one

gaetano-guerriero:
Closing, #1487 supersedes this.

PR #1466, opened 2018-09-15
Check for prompt availability before gathering candidates from shell

galaunay:
# PR Summary
If the shell prompt is not available, trying to gather completion candidates from the python shell will result in emacs hanging (waiting for the shell to be available).

To avoid freezing emacs, elpy now checks for prompt availability before trying to fetch completion candidates.

# PR checklist

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -3544,7 +3544,17 @@ and return the list.
 
   python.el provides completion based on what is currently loaded in the
 python shell interpreter."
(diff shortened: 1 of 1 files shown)


galaunay:
I will try this for a while.
If it prevent completely Emacs from hanging, we may want to set the `elpy-company-add-completion-from-shell` option default value to `t` again.

PR #1476, opened 2018-10-03
Django check settings module by testing exit code instead of output

gaetano-guerriero:
Check for correctness of DJANGO_SETTINGS_MODULE env var was made by ensuring
that python importing wasn't generating any output/error.
This is not safe since during import the settings module can cause logging and
other forms of output.
Checking that python exits 0 status code is more robust and semantically
correct.

# PR Summary

# PR checklist

## For new features only:


jorgenschaefer:
Excellent, thank you!

PR #1479, opened 2018-10-06
Eldoc docstring

galaunay:
# PR Summary
Add the possibility to display a oneline docstring for the symbol at point in eldoc.
Works for modules, classes and functions.

Here is an example of how eldoc works with this feature:

The option `elpy-eldoc-info` replaces `elpy-eldoc-show-current-function` to allow to customize this feature.

# PR checklist

## For new features only:
- [x] An entry in NEWS.rst has been added

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -312,19 +312,6 @@ edited instead. Setting this variable to nil disables this feature."
   :type 'boolean
   :group 'elpy)
 
-(make-obsolete 'elpy-eldoc-show-current-function
-               "Please use `elpy-eldoc-info' instead" "1.25.0")
-
-(defcustom elpy-eldoc-info 'docstring
-  "What to show in eldoc if no calltip is available.
-
-Can be 'docstring' to show a oneline docstring for the symbol at point,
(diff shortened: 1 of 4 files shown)


elpy.el
```diff
@@ -312,6 +312,19 @@ edited instead. Setting this variable to nil disables this feature."
   :type 'boolean
   :group 'elpy)
 
+(make-obsolete 'elpy-eldoc-show-current-function
+               "Please use `elpy-eldoc-info' instead" "1.25.0")
+
+(defcustom elpy-eldoc-info 'docstring
```
jorgenschaefer: I think we can live without this option, can't we? Just use fallbacks, calltip if available, else docstring if available, else current-function?

galaunay:
I used the fallback as you suggested (let me know if I did something weird with the rpc calls).

I also enhanced the algorithm to pick the relevant part of the docstring to display.

I'll test that for a while and merge it if it works fine.

galaunay:
It worked fine so far.

merging !

PR #1482, opened 2018-10-12
Use a more robust way to check for shell availability

galaunay:
# PR Summary
The current way of checking for the shell availability relies on what is currently displayed in the shell buffer, which can be modified and lead to erroneous results. 

Checking if the shell answer quickly enough to a simple statement should be a more robust way of verifying if the shell is available or not.

# PR checklist


gopar:
LGTM

Tests are failing though :P

galaunay:
Glad to see you are back !

Yes, they fail for emacs<24.5, and I can't make it to compile on my machine, which make it a bit difficult to debug...

gopar:
The tests are passing now :)

Is there something else to do before merging this?

galaunay:
I am still not perfectly happy with this...
There should be a better way of checking for shell availability than relying on the last buffer line being a prompt....
But I am running out of ideas.

I wouldn't want to make elpy hangs at every completion attempt because of a weird prompt.

memeplex:
`python-shell-comint-end-of-output-p` is the way python.el does it and, indeed, the way the user visually does it. I know it's not 100% safe but the only alternatives I can figure out are:

1. Somehow pinging the process in a non-blocking way, but this would produce extra prompts or, even if you filter them from the buffer, weird prompt numbering.

2. Launch a thread in to check for some global "I'm ready" variable which I guess doesn't exist at all. Also it's still risky because of the GIL.

I think the current approach is the right one, given the options. Maybe turning off  native completions by default would be a good idea.

galaunay:
I tried to ping the process, but it indeed produces a lot of extra prompts that is very difficult to get rid of.
The native is already turned off by default (#1465).

It is still better than the previous solution, I added a timeout in case it still doesn't work out.
I will make a last review and merge this.

Thanks for your comments.

Issue #1495, opened 2018-11-13
elpy backend error: AttributeError: 'NoneType' object has no attribute 'name'


zhangda82:
# Summary

Thank you for reporting to Elpy !

Here is a template to help you do it so your issue can be fixed as
soon as possible. Please replace this paragraph with a summary of
your issue and fill the next sections.

# Steps to reproduce

# My configuration
## Result of `(elpy-config)`
```
Elpy Error

The backend encountered an unexpected error. This indicates a bug in Elpy. Please open a bug report with the data
below in the Elpy bug tracker:

https://github.com/jorgenschaefer/elpy/issues/new

```
Error Message

'NoneType' object has no attribute 'name'

Configuration

Virtualenv........: None
RPC Python........: 3.6.1 (c:/Anaconda3/pythonw.exe)
Interactive Python: ipython (c:/Anaconda3/Scripts/ipython.exe)
Emacs.............: 26.1
Elpy..............: 1.26.0
Jedi..............: 0.12.1 (0.13.1 available)
Rope..............: 0.9.4-1
Autopep8..........: 1.3.5 (1.4.3 available)
Yapf..............: 0.20.0 (0.24.0 available)
Black.............: Not found (18.9b0 available)
Syntax checker....: flake8.exe (c:/Anaconda3/Scripts/flake8.exe)

Traceback

Traceback (most recent call last):
  File "c:\home\zhangda\.emacs.d\elpa\elpy-20181103.1105\elpy\rpc.py", line 100, in handle_request
    result = method(*params)
  File "c:\home\zhangda\.emacs.d\elpa\elpy-20181103.1105\elpy\server.py", line 79, in rpc_get_oneline_docstring
    get_source(source), offset)
  File "c:\home\zhangda\.emacs.d\elpa\elpy-20181103.1105\elpy\server.py", line 44, in _call_backend
    return meth(*args, **kwargs)
  File "c:\home\zhangda\.emacs.d\elpa\elpy-20181103.1105\elpy\jedibackend.py", line 183, in rpc_get_oneline_docstring
    raw_name = assignment.name
AttributeError: 'NoneType' object has no attribute 'name'
```

```

## Elpy configuration in my init.el
```
# Paste it here
```
(setq elpy-rpc-error-timeout 600000000)
(setq elpy-rpc-timeout 5)
(setq elpy-disable-backend-error-display t)

(defun my-elpy-mode-hook ()
  (highlight-indentation-mode -1)
[...]

PR #1496, opened 2018-11-14
Fix corner case

galaunay:
# PR Summary
Following #1495.

Minor fix of a forgotten if statement in oneline_docstring.

# PR checklist

### DIFF
--- a/elpy/jedibackend.py
+++ b/elpy/jedibackend.py
@@ -179,7 +179,7 @@ class JediBackend(object):
                 raw_name = definition.name
                 name = '{} {}'.format(raw_name, definition.type)
                 doc = definition.docstring().split('\n')
-            elif definition.type in ['instance']:
+            elif definition.type in ['instance'] and assignment not None:
                 raw_name = assignment.name
                 name = raw_name
                 doc = assignment.docstring().split('\n')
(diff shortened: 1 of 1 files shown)


elpy/jedibackend.py
```diff
@@ -179,7 +179,7 @@ def rpc_get_oneline_docstring(self, filename, source, offset):
                 raw_name = definition.name
                 name = '{} {}'.format(raw_name, definition.type)
                 doc = definition.docstring().split('\n')
-            elif definition.type in ['instance']:
+            elif definition.type in ['instance'] and assignment not None:
```
gopar: Wait, isn't there supposed to be an `is` in the statement as well?
```suggestion
            elif definition.type in ['instance'] and assignment is not None:
```

zhangda82: This still triggers an error. 
A quick fix as below worked, although I am not sure it it fits in the context of this function. 
            elif definition.type in ['instance'] and (assignment is not None) and hasattr(assignment, "name"):

galaunay: Sorry for the delay on my side.

Thanks for the fixes.
In my understanding, assignments should always have a `name` attribute.
But better safe than sorry.

PR #1501, opened 2018-11-23
Removed doc of obsolete funs.

rpgoldman:
Removed doc for elpy-use-ipython, elpy-use-cpython.
Also moved refs to ipython so they weren't lost.


gopar:
Thanks!

PR #1503, opened 2018-11-28
Two minor documentation fixes and one enhancement

sten0:
# PR Summary
This is just a minor typo/grammar fix detected by Debian's "lintian" automated testing, plus a one line reflow in elpy.el so that the documentation string matches the documentation.  Please feel free to change it if I misunderstood the intent of the phrase.

# PR checklist

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -54,6 +54,10 @@ this line and may simply install Elpy and all of its recommended
 dependencies with the following command:::
 
   sudo apt install elpa-elpy
+  # To extend Elpy's capabilities by installing packages at the "Suggests"
+  # level of dependency, noted by apt during installation, do:
+  sudo apt install python3-autopep8 yapf3 black python3-pip \
+                   python3-jupyter-console
 
 Elpy can then be activated by running ``M-x elpy-enable``
 
--- a/docs/ide.rst
+++ b/docs/ide.rst
@@ -109,7 +109,7 @@ documentation is a good place for further information.
     If t, add completion candidates gathered from the current python shell.
     Normally elpy provides completion using static code analysis (from jedi).
     With this option set to t, elpy will add the completion candidates from the
-    associated python shell. This allow to have decent completion candidates
+    associated python shell. This provides decent fallback completion candidates
     when the static code analysis fails.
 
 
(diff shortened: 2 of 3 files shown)


sten0:
I decided to add 3f2449aac0493fedbb377de5b9ed2ac0b5a96403 as part of this PR instead of filing a new one.  Explanation is in the commit message.  Please let me know if you'd prefer to NACK it, or if it should be improved in some way.

gopar:
Should the readme advocate users to install the python packages globally?

sten0:
@gopar I'm not sure, and this might be one of those points where Debian culture and best practises is different from Python developer culture and best practises...  I suppose including this in the Elpy project README might implicitly encourage building an application for the Python stack provided in Debian stable (or in the case of Debian testing, for the next stable version), rather than a project-specific mix of versioned dependencies?

One of autopep8, yapf3, or black is needed for maximum user satisfaction, but a lab/tutorial supervisor might not want Intro to Programming students to have access to.  Pip isn't installed by default because the Debian way is to use signed system-wide packages provided by Debian.

More experienced users will of course figure this out on their own, but one of my goals (arguably windmills) is help the new generation of programmers (who learn Python first) have the best chance at an Emacs experience that will inspire them to stick with it; that was my experience/inspiration, on old Sun Ultra workstations, with C++.  I don't know if keeping 3f2449a as a Debian/and_derivatives-specific patch is enough when most new users are probably discovering Elpy with Google, and are probably either using MacOS (Debian instructions don't apply) or Ubuntu (higher than average chance of switching to another application if the first doesn't work out of the box).  Of course, I defer to your wisdom ;-)

jorgenschaefer:
Building Python applications without virtualenvs is a bad idea for all but the most trivial projects. I don't think the distribution model followed by Debian and others works for modern languages much at all anymore. But I'm a nostalgic Debian person myself, so I like the Debian instructions there. I'm just very torn on those.

sten0:
Closing because there are now merge conflicts, 532ce425410a7397a9b5be3611ca15c56623a120 has been solved, 3f2449aac0493fedbb377de5b9ed2ac0b5a96403 doesn't appear to have consensus/approval, and I'm now providing something to this effect in `README.Debian`.

Issue #1507, opened 2018-12-04
Fallback to inferior `help()`


memeplex:
Along the lines of having extended jedi completions with shell completions (in https://github.com/jorgenschaefer/elpy/issues/1399) why not extending `C-c C-d` by calling `help(...)` with the symbol at point as a fallback and only then, in case this fails, prompting the user for an identifier? You can add a customization option like `elpy-doc-show-help-from-shell` or something like that if you don't like the fallback as a default.

This would be useful, for example, if I defined `z = []` in the shell and I'm now running a snippet including `z.append` from a file and want to inspect the documentation for `z.append`. But, also, it could be bound to some key in the inferior mode itself so as to spare the user to type `help(z.append)`.

memeplex:
Also it would be useful in a more important way. Regarding completions, I see that even when the symbol is found by jedi often the shell contributes lots of new completions (see for example `os.`). This means that the shell will often contribute help even to methods and functions in variables already recognized by jedi.

memeplex:
Note: `python.el` already installs `__PYDOC_get_help` but this is designed for eldoc so it only returns an oneliner (there is room for another "fallback to inferior eldoc" here but I'm not requesting it ;) ). I see no problem in directly calling `help()` though, its output hasn't the same format than the one returned by elpy rpc but it's still nice enough for a best effort solution.

galaunay:
I think that's a good idea.
Not too difficult to implement as well.

I put the code checking for the shell availability in a dedicated function in #1482 (`elpy-shell--check-if-shell-available`), so we can use it for this as well.
We can also use the already existing option `elpy-get-info-from-shell` to activate or deactivate this.

galaunay:
I made #1509  to address this.

PR #1509, opened 2018-12-09
Fallback to shell 'help()' for docstring when jedi fails

galaunay:
# PR Summary
Follow #1507.
When jedi fails to provide documentation, use the shell `help()` function to try to get the docstring.

# PR checklist

## For new features only:


galaunay:
Note to myself:
pydoc seems to be confused from time to time with python2.7.
Maybe we should provide this only for python > 3 ?

Issue #1505, opened 2018-12-03
Interpreter shows wrong line number (not the same issue as 983)


steve-the-bayesian:
# Summary
Interpreter shows the wrong line number:

# Steps to reproduce
Here's a function:

```
def exponential_smoothing(x, present_weight):
    ''' Compute an exponential smooth of a data series x.

    Args:
      x:  The series to be smoothed.
      present_weight: The weight to put on the present time period.

    Returns:
      A smoothed series 's' of the same length as x, defined according to

        s[t] = a * x[t] + (1-a) * s[t-1]

    where a is 'present_weight'.
    '''
    s = []
    s[0] = x[0]
    i = 1;
    while np.isnan(s[i - 1]):
        s[i] = x[i];
        i += 1
    for index in range(i, len(x)):
        s[index] = present_weight * x[i] + (1 - present_weight) * s[i - 1]
    return s

```
Here's some test 
import numpy as np
x = np.random.randn(10)
y = exponential_smoothing(x, .01)

The 'def' line of my function begins on line 300 of a larger module.
With elpy turned off  (i.e. with the line (elpy-enable) commented out in my .emacs file) I send my function to the interpreter with C-r.  Then I run the test code, which produces the error message:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/steve/code/python/Rlib/Rlib.py", line 315, in exponential_smoothing
    s[0] = x[0]

With elpy turned on (un-comment (elpy-enable) in .emacs) I send the function to the interpreter by C-return.  Then I send the same test code and get 

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/steve/code/python/Rlib/Rlib.py", line 16, in exponential_smoothing
    Args:
IndexError: list assignment index out of range

It is the same error, but the line number is reported counting from the def line instead of from the top of the file. 

# My configuration
## OS
Ubuntu
No LSB modules are available.                                                    Distributor ID: Ubuntu                                                           Description:    Ubuntu 18.04.1 LTS                                               Release:        18.04                                                            Codename:       bionic
--

## Result of `(elpy-config)`
```
Elpy Configuration

Virtualenv........: None
RPC Python........: 2.7.15 (/usr/bin/python)
Interactive Python: python3 (/usr/bin/python3)
Emacs.............: 25.2.2
Elpy..............: 1.26.0
Jedi..............: 0.13.1
Rope..............: Not found
Autopep8..........: 1.4.3
[...]

PR #1520, opened 2019-01-13
Fix wrong line number in interpreter errors

galaunay:
# PR Summary
Follow #1505.

Sending region of code with Elpy confuses the interpreter, that displays wrong line numbers when error are raised or when using pdb.

Replacing the function extracting region of codes in Elpy (`elpy-shell--region-without-indentation`) with a function from python.el (`python-shell-buffer-substring`) fixes this issue.

# PR checklist


gopar:
LGTM

lihebi:
Hi, I think this change may cause undesired behavior for `elpy-shell-send-statement-and-step`.

Assume we have this code:

```python
def foo():
      1+1
      while True:
               2+2
```

When executing `elpy-shell-send-statement-and-step` on line 2, I expect just `"1+1"` is send to the shell, and I should see the output `2`. Similarly when executing the command on line 4, I expect `"2+2"` is sent and prints out `4`. That is the old behavior.

But now, if the statement under execution is indented, `python-shell-buffer-substring` will wrap the string with `"if True"` and a couple of `\n`, thus becoming `if True:\n\n   1+1`, where the number of `\n` is the indentation level of the statement.

The downsides of this new behavior:
1. It does not print out the result of the statement, because it is wrapped with the conditional stmt
2. There can be a lot of empty lines inserted into the shell REPL, thus pretty annoying, like this:

```
In [7]: if True:
   ...: 
   ...: 
   ...: 
   ...: 
   ...:             2+2
In [8]: 
```

PR #1534, opened 2019-02-16
Nicer display of inputs echoing

galaunay:
# PR Summary
Follow second discussion in #1505.

## Problem
For this given code:
```Python
def foo():
      1+1
      while True:
               2+2   # <- cursor here
```
Following commit #1520, using `elpy-shell-send-statement-and-step` would lead to the following result in the shell buffer:
```python
In [7]: if True:
   ...: 
   ...: 
   ...: 
   ...: 
   ...:             2+2
In [8]: 
```
which is quite ugly and doesn't echo the excepted output value `4`.

## Solution
Filtering the `if True:` and removing the indentation on the region lead to the much nicer echo.
Slightly modifying the function `elpy-shell-send-file` allows to get the output back
```Python
In [7]: 2+2  
Out[7]: 4     
In [8]: 
```

## 

# PR checklist

### DIFF
--- a/docs/ide.rst
+++ b/docs/ide.rst
@@ -477,14 +477,6 @@ evaluation command, thereby providing visual feedback.
    value). Then, no window is needed to display the shell (thereby saving screen
    real estate), but the outputs can still be seen in the echo area.
 
-.. option:: elpy-shell-capture-last-multiline-output
-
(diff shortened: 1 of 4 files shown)


gopar:
LGTM

galaunay:
It was a bit tricky.
I did as much tests as possible, let's hope it is not gonna break anything :/

PR #1537, opened 2019-02-18
WIP: Add Debugging features

galaunay:
# PR Summary
Add functions to ease debugging.

- `elpy-pdb-toggle-breakpoint-at-point`: set (or unset) a breakpoint at the current line.
- `elpy-pdb-clear-breakpoints`: clear all breakpoint in the buffer.
- `elpy-pdb-debug-buffer`: start pdb on the current buffer
- `elpy-pdb-break-at-point`: start pdb on the current buffer and jump to the current line
- `elpy-pdb-debug-last-exception`: start post mortem pdb on the last exception

## Todo:
- [x] Add functions
- [x] Add ability to set breakpoints at specific lines, with fringe indicator
- [x] Add keybindings
- [x] add menubar entries
- [x] Add documentation
- [x] Add tests


gopar:
Ohhhhhhh niiiiiice!!!!!
Had no idea pdb could do post mortem. Guess that's what happens when you don't read the docs :P 

Oh and it LGTM

gopar:
What other funcs are you planning on adding?
I noticed that you have it labels as WIP so just curious :P

galaunay:
Glad you like it :).

I was thinking about the possibility of setting breakpoints to specific lines (with a nice fringe indicator).
(I added that to the todo list).

It may worth looking at what other IDE do to get some additional ideas.

galaunay:
Finally done.

usnigg:
Thanks a lot for this feature! There is one problem I ran into and maybe someone with more experience could help with:
I use ipython 7.4.0 with python 3.7.3.
When I set a breakpoint at some line using C-c C-g b and then run the script it nicely enters the pdb shell and stops at that line. However when I either quit the debugger (q) or continue through the end (c), I can then no longer use the ipython buffer, i.e. if I type a valid python command e.g. '5' I get:

KeyError                                  Traceback (most recent call last)
~/Dev/memory/hopfield/hopfield.py in <module>
----> 1 5

~/.virtualenvs/memory/lib/python3.7/site-packages/IPython/core/displayhook.py in __call__(self, result)
    256             self.write_output_prompt()
    257             format_dict, md_dict = self.compute_format_data(result)
--> 258             self.update_user_ns(result)
    259             self.fill_exec_result(result)
    260             if format_dict:

~/.virtualenvs/memory/lib/python3.7/site-packages/IPython/core/displayhook.py in update_user_ns(self, result)
    194 
    195         # Avoid recursive reference when displaying _oh/Out
--> 196         if result is not self.shell.user_ns['_oh']:
    197             if len(self.shell.user_ns['_oh']) >= self.cache_size and self.do_full_cache:
    198                 self.cull_cache()

KeyError: '_oh'

I have looked a little bit online for more information and it seems this may have something to do with: https://github.com/ipython/ipython/issues/136 although perhaps not exactly the same issue.

I think it's a ipython specific problem since it works normally when I use the python interpreter shell instead.

I think it should be possible to fix this because if I set the breakpoint instead "manually" with 'import pdb; pdb.set_trace()' then everything works fine, i.e. after exiting the pdb shell the ipython shell is still useable.

Here is a minimal script to try and reproduce the problem:

def f(x):
    y = x + 1


if __name__ == '__main__':
    f(4)

And here is the workflow that yields the error:
1) start an ipython shell from elpy with C-c C-z
2) set a breakpoint at the line 'y=x+1' inside of f(x) --> a nice read circle appears in the sidebar
3) Execute C-c C-g g --> the ipython buffer is 'replaced' by a pdb shell stopped at the line with the breakpoint
4) (optional) use pdb commands such as 'n' etc.
5) quit the debugger with 'q'
6) type any valid command at the ipython prompt --> the error copied above is raised

Any suggestions for how to fix this appreciated. Thanks!

galaunay:
Hi, 

I could not reproduce this with IPython on mine.

Could you create a [new issue](https://github.com/jorgenschaefer/elpy/issues/new) with your Elpy configuration and OS so I can investigate a bit further ?

Issue #1546, opened 2019-02-25
Doesn't work inside a virtualenv


raxod502:
# Summary

Autocompletion works as desired when I _don't_ activate a virtualenv, as I've installed `jedi` globally. However, as soon as I _do_ activate virtualenv, autocompletion suddenly stops working at all, as I do not have `jedi` installed in my virtualenv (it's for a shared project managed via Pipenv).

I can see why this happens—if Elpy tries to run `jedi` via whatever `python` is available on the PATH (which would be the one from my virtualenv) then of course it won't be able to find `jedi`. But this behavior seems like a bug to me. Why isn't Elpy running `jedi` through the global `python`?

I don't know how to work around this except by installing all of my project dependencies globally or by installing all of Elpy's dependencies into my project's `Pipfile`, both of which are totally horrifying options.

It seems like #1020 is reporting the same problem but no solution is presented (unless there is some way to automatically install extra packages into every virtualenv without interfering with anything?).

# Steps to reproduce

* `M-x pyvenv-workon`
* Try to use autocompletion

# My configuration
## OS

macOS 10.14.3

## Result of `(elpy-config)`

_Before_ activating virtualenv:

```
Virtualenv........: backend-hrC0GD4S (/Users/raxod502/.virtualenvs/backend-hrC0GD4S)
RPC Python........: 3.7.2 (/usr/local/bin/python3)
Interactive Python: python3 (/usr/local/bin/python3)
Emacs.............: 27.0.50
Elpy..............: 1.26.0
Jedi..............: 0.13.3
Rope..............: 0.12.0
Autopep8..........: 1.4.3
Yapf..............: 0.26.0
Black.............: 18.9b0
Syntax checker....: flake8 (/Users/raxod502/.local/bin/flake8)
```

_After_ activating virtualenv:

```
Virtualenv........: backend-hrC0GD4S (/Users/raxod502/.virtualenvs/backend-hrC0GD4S)
RPC Python........: 3.7.2 (/Users/raxod502/.virtualenvs/backend-hrC0GD4S/bin/python3)
Interactive Python: python3 (/Users/raxod502/.virtualenvs/backend-hrC0GD4S/bin/python3)
Emacs.............: 27.0.50
Elpy..............: 1.26.0
Jedi..............: Not found (0.13.3 available)
Rope..............: Not found (0.12.0 available)
Autopep8..........: Not found (1.4.3 available)
Yapf..............: Not found (0.26.0 available)
[...]

PR #1548, opened 2019-02-25
Use a dedicated virtualenv for the RPC

galaunay:
# PR Summary
Follow #1546.

At the moment, when using a virtualenv, jedi (and all the other python packages used by the IDE) need to be installed in the virtualenv to work.

This can be inconvenient as one would need to install jedi (and the others) in every virtualenv.

This PR intends to start the rpc in a dedicated virtualenv.
It makes easier to install Elpy's python dependencies (jedi, yapf, rope, ...) and save the burden for the user to install them manually in every virtualenv.
The changement should not break any user workflow.

# Todo/note list

- [x] Make Elpy start the rpc in a dedicated virtualenv
- [x] Update `elpy-config` command to check for packages in the dedicated virtualenvs
- [x] Update button in `elpy-config` to install needed package
- [x] Make sure python version mismatch (2.7 vs 3.X) are not a problem
- [x] Update Elpy's `setup` script
- [x] Update the documentation
- [x] Add tests to cover this
- [x] Make sure it doesn't slow down Elpy
- [x] Add a word in the doc about updating the RPC packages


elpy.el
```diff
@@ -1259,7 +1262,12 @@ PyPI, or nil if that's VERSION."
 
 (defun elpy-insert--pip-button-action (widget &optional _event)
   "The :action option for the pip button widget."
-  (async-shell-command (widget-get widget :command)))
+  (with-pyvenv-deactivated
+   (sudo-async-shell-command (widget-get widget :command))))
+
+(defun sudo-async-shell-command (command)
+  (let ((default-directory "/sudo::"))
+    (async-shell-command command)))
```
gopar: Why do we need sudo?
Also, seems it would only work with linux machines

gopar:
One thing that came to mind is that what if the global python is 2.x, but virtualenv is 3.x? 

I'm scared this will add some complexity to the project, and sadly I'm not that active to help out with the burden :(

I tested it out for a bit, but couldn't get it working. (It was rather quick, only debugged for a few minutes)

If this goes forward, wouldn't this break current elpy users? Since I'm assuming most have jedi installed in the virtualenv since this is how it's been working for a while :P 
Also, it doesn't seem to fallback to virtualenv jedi right?

elpy.el
galaunay: Admin rights would be needed to install global packages from Emacs (with the buttons in `elpy-config`).
Thanks for pointing this out, I didn't thought about other OS...
We may have to drop the `elpy-config` buttons in that case.

galaunay:
> One thing that came to mind is that what if the global python is 2.x, but virtualenv is 3.x?

Actually, it should not be a problem. It is not perfectly clear from the documentation, but Jedi should understand both 2.x and 3.x code, regardless of the python version it is started from. (Quickly tested today between 2.7 and 3.6). I will add that to the todo list to be sure not to forget about this point.

> I tested it out for a bit, but couldn't get it working. (It was rather quick, only debugged for a few minutes)

Unfortunately, I cannot say it is a surprise :). 
But as the current version is working for me, it would be very valuable to know why it is not for you.

> If this goes forward, wouldn't this break current elpy users?

That is a good point, and we have to make sure it will be perfectly transparent for the current Elpy users.
That is why I wanted to fall back to using jedi from the virtualenv if it was not installed globally.
(Which is not implemented yet, you are right).

I don't see this PR as a priority right now, as it is quite a dangerous one and will need to carried out very carefully. I will just try to work on it when I can. 
I may need your feedback at some point though :).

aaron-kaplan:
I share the desire not to have to install jedi et. al. in every virtualenv, but I wouldn't want to depend on the system python to provide it either. The system python belongs to the distro maintainers. If you install packages into it using the distro's package manager (e.g. `apt`) then you have to settle for whatever version they decided to package, and if you install the versions you want using `pip` then you risk destabilizing the system.

I haven't understood all the changes in the PR yet, but perhaps we could install jedi et al. into a special virtualenv dedicated to elpy dependencies, rather than into the system python?

galaunay:
@gopar, @aaron-kaplan, @raxod502, @PeterChristoph.

I made this PR to have the RPC running in a dedicated virtualenv (thanks @aaron-kaplan  for the idea).
It means that you don't need to install any additional python packages (jedi, yapf, rope, ...) in your virtualenv anymore.

It should be completely transparent for the current users an simplify the installation of Elpy for new users.

However, I had to make quite a lot of modifications to the RPC, and even with the additional tests, I am not sure It will work seamlessly for any configuration.
It would be very useful if some of you could try the PR and tell me if something went wrong.

Thanks, and hopefully I will be able to merge this soon.

raxod502:
That looks great! Although I will say that I personally think LSP is the future, and so I've migrated my configuration to use it and have set up automatic detection&activation of Poetry and Pipenv virtualenvs when starting the LSP server:

* https://github.com/raxod502/radian/blob/7ca1c1aa6c9605d1ef456bb94803bdfb4698c005/emacs/radian.el#L3190-L3213
* https://github.com/raxod502/radian/blob/7ca1c1aa6c9605d1ef456bb94803bdfb4698c005/emacs/radian.el#L3234-L3241

saiduc:
Has there been any progress on merging this feature? I am looking forward to it being part of elpy

galaunay:
@raxod, I will give LSP another try as soon as soon as I can.
Last time I tried to use LSP, it was not allowing to replace all of Elpy's features.
Nothing really important, but I don't want to break anybody workflow...

@saiduc, It is now working and properly tested.
I just wanted to run some additional tests to make sure it will be completely transparent to current users. If everything works fine, I intend to merge it this week.
Have you tried it ? any feedback ?

galaunay:
Note to myself: need to update the melpa recipe to include `requirements-rpc` files.

vindarel:
Great addition, thank you!

PR #1563, opened 2019-04-05
Clean up tests under py2.7

dickmao:
Several tests failed when I ran `./scripts/test` from the command line under py2.7.


galaunay:
Lgtm, thanks, 

I just don't get what was the problem with `elpy-promise-wait-should-return-early-for-resolved-promise`.
Is it taking more time than expected for you ?

dickmao:
> I just don't get what was the problem with `elpy-promise-wait-should-return-early-for-resolved-promise`.
> Is it taking more time than expected for you ?

Yes.  When the margin for error is a tenth of a second, the test fails every time out of the box.  I suppose if I ran it enough times, it might succeed.

When I stretched the margin for error to 9/10 of a second, the test succeeds every time.

galaunay:
I just resolved the conflict, thanks for fixing that.

Issue #1584, opened 2019-04-20
elpy-config is extremely slow


mbtuckersimm:
# Summary

`elpy-config` is extremely slow, to the point where I've given up. I've done some investigation (detailed below) and it seems that the issue is in contacting the pypi server, despite the fact that I have a quite fast internet connection.

Even though the underlying problem is likely not with `elpy`, I'm stuck as to how to fix it, so I'm filing this issue and hoping for some help.

# Steps to reproduce

1. Load a file in a python project.
2. Activate virtualenv with `M-x pyvenv-activate`.
3. `M-x elpy-config`
4. Wait... 

# My configuration

OS: Ubuntu 16.04
Emacs: 26.1
elpy: 20190416.1701 (melpa)

# Result of `(elpy-config)`
Unavailable since `elpy-config` never completes (well, I run out of patience).

## Elpy configuration in my init.el
```
(use-package elpy
  :config
  (elpy-enable)
  (setenv "WORKON_HOME" "~/.pyenv/versions/")
  ;; disable flymake since we use flycheck instead
  (setq elpy-modules (delq 'elpy-module-flymake elpy-modules))
  :ensure t)
```

# What I've tried so far

I was playing around in an ipython shell (within the relevant virtual environment) with the python code that's called during `elpy-config`:
1. Import `json`, `sys`, and `warnings`.
2. `import xmlrpc.client as xmlrpclib`
3. Define the function `latest`:
```
def latest(package, version=None):
    try:
        with xmlrpclib.ServerProxy('https://pypi.python.org/pypi') as pypi:
            latest = pypi.package_releases(package)[0]
        if version is None or LooseVersion(version) < LooseVersion(latest):
            return latest
        else:
            return None
    except:
        return None
```
4. `import timeit`
5. `timeit.timeit("latest('jedi')", setup='from __main__ import latest', number=1)`
6. Result is `524.0016240759978`.

So it takes almost 9 minutes to do this operation for just a single package! (The `latest` function does come up with the correct latest version of jedi, 0.13.3 at the time of filing this issue, FWIW.)

mbtuckersimm:
I should mention that this is on a desktop computer, so it's not easy to try this with a different internet connection :(

gopar:
Yups, I see the same results. Might be better to switch to the new json api that pypi provides. Now that warehouse(New version of pypi) is live

Note to self: 
Warehouse docs: https://warehouse.pypa.io/api-reference/json/

[...]

PR #1585, opened 2019-04-21
Replace xmlrpclib with urllib2

gopar:
# PR Summary
Replace `xmlrpclib` with `urllib2`.

There has been issues with xmlrpclib being slow when doing
`elpy-config`. Luckily, with the new version of pypi, there are some
new json API's that can be used. These changes reflect that.

I noticed that SSL in Python 3.6+ is an issue on Mac OS when using urllib. (See here: https://stackoverflow.com/a/42334357/2178312)
Something to be aware of when users come to elpy about this but then again, this isn't elpy specific issue 🤷‍♂️ 

# PR checklist

## For new features only:
- [N/A] Tests has been added to cover the change
- [N/A] The documentation has been updated

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -673,17 +673,17 @@ import warnings
 warnings.filterwarnings('ignore', category=FutureWarning)
 
 try:
-    import xmlrpclib
+    import urllib2 as urllib
 except ImportError:
-    import xmlrpc.client as xmlrpclib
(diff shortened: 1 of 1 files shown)


gopar:
Should solve #1584

galaunay:
Nice, 
quite fast indeed !

Issue #1596, opened 2019-05-07
Downstream user requested QuickStart [Debian bug #927085]


sten0:
Hi,

A Debian user filed a bug requesting a QuickStart doc, either as a section or standalone (at any rate, probably docs/quickstart.rst).  User appears to be coming from a Python, but not Emacs background.  It might also be nice to provide a corresponding docs/refcard.rst for users of similar background.  Maybe a refcard.rst with the most used shortcuts/functions `highlighted` would do the trick?

1) User requests a short list of the most used functions and their shortcuts.
   * User requests highlighting of `elpy-shell-send-region-or-buffer` in particular
   * docs/ide.rst already provides this info in comprehensive form (P.S. that's what I prefer, thank you!)
2) User requests an `add-hook` emacs.el hint for `elpy-shell-switch-to-shell`.

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=927085

Thanks!
Nicholas

galaunay:
Hi, 
Thanks for forwarding this.

I added a `quickstart` section in the doc in PR #1598 (any feedback on that is welcome).
The documentation on readthedoc is stuck to elpy 1.27.0 for now, so it may take a bit of time to be updated.

About the hook, I guess you are refering to this:
> Do you happen to know if it is possible to configure emacs to open a
Python buffer in elpy mode by default and skip the C-c C-z?

My answer would be that you don't need to open the shell with `C-c C-z` before evaluating the code with `C-c C-c`. `C-c C-c` will automatically open a python shell if needed. you can still open a shell automatically when opening a python file with `(add-hook 'elpy-mode-hook 'elpy-shell-display-buffer t)`

> 6) Your bug report template should include the output of
> elpy-config. I've included mine at the bottom. I'm not sure what is going on with
>
>    Elpy..............: Not found (Python), 1.28.0 (Emacs Lisp)
>
>though. What's the difference between those two cases?

This had been fixed in Elpy 1.29.0.

> > 3) I suggest mentioning the upstream manual in README.Debian. Namely,
> > https://elpy.readthedocs.io/en/latest/
> >
> 
> Unfortunately this will not provide accurate documentation for users
> of Debian buster in the coming months, because that URL will document
> a newer version of Elpy, and buster is shipping with 1.28.0.  Also, a
> link is useless for users who are offline.

I think we could provide documentation for all Elpy versions on readthedoc, so you could link to the version of the doc debian is actually shipping.
Let me know if it would be useful.

[...]

PR #1602, opened 2019-05-21
Enable elpy-mode in python buffer when enabling Elpy

galaunay:
# PR Summary
Related to #1596.

If Elpy is enabled (with `elpy-enable`) while some python files are already open, `elpy-mode` will not be activated in them (before `elpy-mode` uses `python-mode-hook`).

This PR makes sure that elpy-mode is properly activated in already existing python buffer.

# PR checklist


gopar:
LGTM

Issue #1606, opened 2019-05-29
wiki overlaps with documentation and README


aspiers:
Hi there, it's a really nice problem to have too much documentation ;-)  But I would like to suggest that the wiki documentation and much of the README.md should be merged into `docs/`, and then the wiki retired, because:

- Having more than one source means they will inevitably diverge and cause confusion.  For newcomers (like me) don't know whether to follow https://github.com/jorgenschaefer/elpy/blob/master/README.rst or https://elpy.readthedocs.io/en/latest/introduction.html#installation or https://github.com/jorgenschaefer/elpy/wiki/Installation.
- When code changes impact the documentation, the docs can be updated in the same commit as the code, ensuring they stay in sync.
- The wiki is not subject to peer review, therefore harder to maintain quality.

Is this something which you could agree to in principle, even if you don't have time to do it?  If so, I might be able to contribute a few PRs in this direction.

aspiers:
Just noticed that the README also covers installation, so this is three different places to choose from.  Updated the issue accordingly.

gopar:
Hey,

Thanks for bringing this up :)

I see your point, and I agree. Personally, I would prefer that the README still have a quick "Get Started Guide" where it shows the bare minimum to get elpy running, and then link to the docs for everything else.

As for the wiki, I believe some stuff there is outdated, perhaps move some of the sections to the docs.

@galaunay Thoughts about moving mostly everything to https://elpy.readthedocs.io/?

I don't know anything about how https://elpy.readthedocs.io/ is updated, or if it can simply be updated by editings the `rst` files. :P

galaunay:
@aspiers I do agree that the wiki should be removed (I just read through it quickly and there is a lof of obsolete information in there). I think the relevant information could be perfectly merged in the docs as you suggest. 
I can find some time to work on that, but if you are willing to contribute, that would help a lot.

@gopar I find as well that the readme could be more appealing and reflect better what is Elpy. A quick start section would be a very good idea. I was also thinking about a having a short list of features (with links to the doc). Would it be only to show that Elpy can offer the same kind of features (sometimes more ?) than other IDEs.

> I don't know anything about how https://elpy.readthedocs.io/ is updated, or if it can simply be updated by editings the rst files. :P

It will, as soon as I can get in touch with Jorgen and get access to Elpy's readthedocs account :).

aspiers:
Awesome, thanks folks!

[...]

PR #1608, opened 2019-06-03
Reformat WIki, readme and documentation

galaunay:
# PR Summary
Follow #1606.
Attempt to refresh and enhance Elpy's documentation.

## Todo list:
- [x] Reformat readme
- [x] Merge the wiki in the docs
- [x] Delete the wiki
- [x] Update (and reformat ?) the docs
- [x] Complete the doc for missing features (debugging, elpy's snippets, others ?) 
- [ ] Update readthedocs configuration
- [x] Test modifications format and links on readthedocs (for FAQ and tips)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -36,9 +36,9 @@ Features
 - `On-the-fly syntax checking`_
 - `Access to documentation`_
 - `Debugging`_
-- `Test running`_
+- `Testing`_
 - `Profiling`_
-- Snippet Expansion (using `yasnippet`_)
+- `Snippet Expansion`_
 - Code hinting
 
 .. _Code completion: https://elpy.readthedocs.io/en/latest/ide.html#completion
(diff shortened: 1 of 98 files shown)


README.rst
```diff
@@ -23,34 +16,47 @@ Emacs Lisp as well as Python.
...
-  pip install yapf
-  pip install black
 
-One-line install: ``pip install jedi flake8 autopep8 black yapf``
+Installation
+============
 
-With `use-package`
-------------------
+The most straightforward way of installing elpy is to use `use-package`:
```
sten0: @galaunay for now ;-)  If the Debian package added a method to hook into python-mode (as-needed), `apt` would install an Elpy that was fully enabled and integrated out-of-the-box.  I'm also considering changing the Debian installation instructions to use the older `apt-get`, because then the `--install-suggests` can be added to the command, which will install all the (optional) Python deps.  IIRC our venvwrapper also defaults to `togglesitepackages`, but I'll need to double check and get back to you on  that point :-)

README.rst
```diff
@@ -59,54 +65,30 @@ With `use-package`
...
 
-Or if you want to defer Elpy loading:
-
-.. code-block:: elisp
-
-  (use-package elpy
-    :ensure t
-    :defer t
-    :init
-    (advice-add 'python-mode :before 'elpy-enable))
```
sten0: Why was this deferred loading example dropped?

galaunay: I though it was not important enough to be in the readme.
It is in the doc though ([here](https://github.com/galaunay/elpy/blob/Reformat-documentation/docs/introduction.rst#with-use-package)).

README.rst
galaunay: That would be nice indeed to have just to install one package to get Elpy to work !
I would be happy to update the doc when it's working.

PR #1623, opened 2019-07-16
Fixed bug where elpy modifies company-frontends global value

TommyX12:
# PR Summary

I noticed that elpy removes `company-echo-metadata-frontend` from the global value of `company-frontends`. I believe this is caused by trying to disable `company-echo-metadata-frontend` to avoid conflicting with eldoc, but `delq` modifies the references to cons cells, thus even though `company-frontends` is made local, its global value still gets affected.

This PR fixes that.

# PR checklist

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -3736,8 +3736,7 @@ and return the list."
      (eldoc-add-command-completions "company-")
      (eldoc-add-command-completions "python-indent-dedent-line-backspace")
      (set (make-local-variable 'company-frontends)
-          (delq 'company-echo-metadata-frontend
-                (mapcar #'identity company-frontends)))
+          (remq 'company-echo-metadata-frontend company-frontends))
      (set (make-local-variable 'eldoc-documentation-function)
           'elpy-eldoc-documentation)
      (eldoc-mode 1))
--- a/elpy.el
+++ b/elpy.el
@@ -3736,7 +3736,8 @@ and return the list."
      (eldoc-add-command-completions "company-")
      (eldoc-add-command-completions "python-indent-dedent-line-backspace")
      (set (make-local-variable 'company-frontends)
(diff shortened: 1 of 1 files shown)


galaunay:
Thanks for this.

Tests are failing for another reason.
I will merge this PR as soon as they are fixed.

galaunay:
Tests should pass properly now.


However, I think i would find the following a bit clearer (in case we need to modify this piece of code at some point in the future):
```Elisp
     (make-local-variable 'company-frontends)
     (setq company-frontends
           (delq 'company-echo-metadata-frontend company-frontends))
```

TommyX12:
> ```emacs-lisp
>      (make-local-variable 'company-frontends)
>      (setq company-frontends
>            (delq 'company-echo-metadata-frontend company-frontends))
> ```

Did you mean
```emacs-lisp
     (make-local-variable 'company-frontends)
     (setq company-frontends
           (delq 'company-echo-metadata-frontend (mapcar #'identity company-frontends)))
```
? :)

galaunay:
Ah, yes you are right, I was confused :/

Anyway, maybe using `remq` instead of `delq` would be a better idea:
```elisp
(set (make-local-variable 'company-frontends)
     (remq 'company-echo-metadata-frontend company-frontends))
```
It would avoid side-effects.
Does it seem sensible ?

TommyX12:
> Ah, yes you are right, I was confused :/
> 
> Anyway, maybe using `remq` instead of `delq` would be a better idea:
> 
> ```emacs-lisp
> (set (make-local-variable 'company-frontends)
>      (remq 'company-echo-metadata-frontend company-frontends))
> ```
> 
> It would avoid side-effects.
> Does it seem sensible ?

Sounds good, fixed!

gopar:
Merged. Thanks!

PR #1630, opened 2019-08-06
Use a decent default for django settings path

galaunay:
# PR Summary
When using Django, Elpy needs the user to set the environment variable `DJANGO_SETTINGS_MODULE` to know where the setting file is.
This can be cumbersome, as the user will have to change this environment variable every time he switches projects.

With this PR, Elpy will use the default setting path (gathered from `manage.py`) if `DJANGO_SETTINGS_MODULE` is not set.
 

# PR checklist


galaunay:
@gopar, I tried to make Django slightly easier to use in Elpy with this PR.
But don't use Django a lot, so I am not sure I did it right...
Could you give me you opinion ?

Issue #1550, opened 2019-02-27
elpy-shell-send-region-or-buffer show strange letters ^G


urugang:
# Summary
I cannot use C-c C-c to send region to shell.

# Steps to reproduce
test.py 
```
alist = [i for i in range(3)]
numbers = [-1, -2, -3]
```
1) elpy-shell-switch-to-shell
2) python-shell-send-region 
it works  as expected.
3) elpy-shell-send-region-or-buffer
shell pended as beblow:
```
In [14]: alist = [i for i in range(3)]
    ...: numbers = [-1, -2, -3]
^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G^G

```

# My configuration
## OS
macos 10.14.2

## Result of `(elpy-config)`
```
Virtualenv........: None
RPC Python........: 3.7.2 (/usr/local/bin/python3)
Interactive Python: jupyter (/usr/local/bin/jupyter)
Emacs.............: 26.1
Elpy..............: 1.28.0
Jedi..............: 0.13.3
Rope..............: 0.12.0
Autopep8..........: 1.4.3
Yapf..............: 0.26.0
Black.............: Not found (18.9b0 available)
Syntax checker....: flake8 (/Users/urugang/Library/Python/3.7/bin/flake8)

```

## Elpy configuration in my init.el
```
(use-package elpy
  :ensure t
  :config (setq python-shell-interpreter "python3")

  (put 'set-goal-column 'disabled nil)
  (elpy-enable)
  (setq elpy-rpc-ignored-buffer-size (* 1024 1024 1024))


  ;; interactive python
  (setq python-shell-interpreter-args "-i"
        elpy-rpc-python-command "python3")

  (setq python-shell-interpreter "jupyter"
        python-shell-interpreter-args "console --simple-prompt"
        python-shell-prompt-detect-failure-warning nil)
  (setq python-indent-guess-indent-offset t)
  (setq python-indent-guess-indent-offset-verbose nil)
  (add-to-list 'python-shell-completion-native-disabled-interpreters
             "jupyter"))
```

rpgoldman:
[...]

PR #1671, opened 2019-09-25
Fix 1550: Minimize length of string sent to process-send-string

eihli:
# PR Summary

Too long of a string causes the string to be sent in batches
due to the limit of the input buffer size of processes
on some machines.

https://github.com/jorgenschaefer/elpy/issues/1550

This reverts commit 740b0a046f8b1750d8fccd8e578df67efc04e927
and introduces an alternative fix for the performance
enhancement that commit introduced.

The notes and repro details below are copy/pasted from the 1550 issue thread. 

I can recreate this on-demand by changing the length of the string sent to `process-send-string`.

```
DEFUN ("process-send-string", Fprocess_send_string, Sprocess_send_string,
       2, 2, 0,
       doc: /* Send PROCESS the contents of STRING as input.
PROCESS may be a process, a buffer, the name of a process or buffer, or
nil, indicating the current buffer's process.
If STRING is larger than the input buffer of the process (the length
of which depends on the process connection type and the operating
system), it is sent in several bunches.  This may happen even for
shorter strings.  Output from processes can arrive in between bunches.

If PROCESS is a non-blocking network process that hasn't been fully
set up yet, this function will block until socket setup has completed.  */)
```

On my mac, Mojave 10.14.6 and Python 2.7.11, if I checkout `740b0a0~1` and make the following change, then I get the error.

```
diff --git a/elpy-shell.el b/elpy-shell.el
index d16baaf..1ce3e5d 100644
--- a/elpy-shell.el
+++ b/elpy-shell.el
@@ -628,6 +628,14 @@ print) the output of the last expression."
       (concat
        "import sys, codecs, os, ast;"
        "__pyfile = codecs.open('''%s''', encoding='''%s''');"
+	   "('abcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefg)"
+	   "('abcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefg)"
+	   "('abcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefg)"
+	   "('abcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefg)"
+	   "('abcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefg)"
+	   "('abcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefg)"
+	   "('abcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefg)"
+	   "('abcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefg)"
        "__code = __pyfile.read().encode('''%s''');"
        "__pyfile.close();"
        (when (and delete temp-file-name)
```

With that change, this is the error I see when I try to evaulate a buffer.

```
>>> print("HI")
...: print("BYE")
import codecs, os, ast;__pyfile = codecs.open('''/var/folders/jy/nwdc6ngd1012k8mqbb_3tz_c0000gn/T/pyoNfgqv''', encoding='''utf-8''');('abcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefg)('abcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefg)('abcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefg)('abcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefg)('abcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefg)('abcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefg)('abcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefg)('abcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefgabcdefg)__code = __pyfile.read().encode('''utf-8''');__pyfile.close();os.remove('''/var/folders/jy/nwdc6ngd1012k8mqbb_3tz_c0000gn/T/pyoNfgqv''');__block = ast.parse(__code, '''/tmp/foo.py''', mode='exec');__block.body = __block.body if not isinstance(__block.body[0], ast.If) else __block.body if not isinstance(__block.body[0].test, ast.Name) else __block.body if not __
^G
^G
^G
```

Scroll to the right of that code block and you'll notice the string is truncated.

If I remove those lines that do nothing but increase the length of the string, I get the following output.

```
>>> print("HI")
...: print("BYE")
import codecs, os, ast;__pyfile = codecs.open('''/var/folders/jy/nwdc6ngd1012k8mqbb_3tz_c0000gn/T/pypUfhiS''', encoding='''utf-8''');__code = __pyfile.read().encode('''utf-8''');__pyfile.close();os.remove('''/var/folders/jy/nwdc6ngd1012k8mqbb_3tz_c0000gn/T/pypUfhiS''');__block = ast.parse(__code, '''/tmp/foo.py''', mode='exec');__block.body = __block.body if not isinstance(__block.body[0], ast.If) else __block.body if not isinstance(__block.body[0].test, ast.Name) else __block.body if not __block.body[0].test.id == 'True' else __block.body[0].body;__last = __block.body[-1];__isexpr = isinstance(__last,ast.Expr);_ = __block.body.pop() if __isexpr else None;exec(compile(__block, '''/tmp/foo.py''', mode='exec'));eval(compile(ast.Expression(__last.value), '''/tmp/foo.py''', mode='eval')) if __isexpr else None
HI
BYE
>>> 
```

That string is not truncated and my comint buffer doesn't break. I'm still curious why the `import codecs` line is being output. But at least the comint buffer doesn't break and start showing those `^G` symbols.

# PR checklist

## For new features only:

### DIFF
--- a/elpy-shell.el
+++ b/elpy-shell.el
@@ -263,7 +263,9 @@ If ASK-FOR-EACH-ONE is non-nil, ask before killing each python process."
 
 If SIT is non-nil, sit for that many seconds after creating a
 Python process. This allows the process to start up."
-  (let* ((bufname (format "*%s*" (python-shell-get-process-name nil)))
+  (let* ((process-connection-type
+          (if (string-equal system-type "darwin") nil t))  ;; see https://github.com/jorgenschaefer/elpy/pull/1671
+         (bufname (format "*%s*" (python-shell-get-process-name nil)))
          (proc (get-buffer-process bufname)))
     (if proc
         proc
--- a/test/elpy-module-yasnippet-test.el
+++ b/test/elpy-module-yasnippet-test.el
@@ -34,8 +34,8 @@
 
 (ert-deftest elpy-snippet-split-args ()
   (elpy-testcase ()
+    (elpy-enable)
     (python-mode)
-    (elpy-mode)
     (let ((args "self, arg1, arg2 :int, arg3=4, arg4='message', arg5 :int, arg6 :float=65.4, *args, **kwargs"))
       (should (equal (elpy-snippet-split-args args)
 	     '("self" "arg1" "arg2" "arg3" "arg4" "arg5" "arg6" "*args" "**kwargs"))))
--- a/test/elpy-shell-send-file-test.el
+++ b/test/elpy-shell-send-file-test.el
@@ -1,0 +1,11 @@
+(ert-deftest elpy-shell-send-file-should-accept-large-strings ()
+  (elpy-testcase ()
+   (elpy-enable)
+   (python-mode)
+   (let ((test-string (make-string 6000 ?a))
+         (process (elpy-shell-get-or-create-process)))
+     (python-shell-send-string
+      (format "print('start');foo='%s';print('end');" test-string) process)
+     (with-current-buffer (process-buffer process)
+       (elpy/wait-for-output "start")
+       (should (string-match "end" (buffer-string)))))))


galaunay:
Sorry to have taken so long to reply.

I am a bit confused about the caching you implemented, specially by this [line](https://github.com/jorgenschaefer/elpy/pull/1671/files#diff-530aac63f11f253c055bc4fc83d77d06R153):
```
  (if (not (eq elpy-shell--python-version-cache python-shell-interpreter))
```
If I understand properly, `elpy-shell--python-version-cache` is holding the python version ("3.7.4" for example), while `python-shell-interpreter` is holding the name of the python binaries ("python", or "python2").
Is the `eq` test not always `nil` then ?

eihli:
Yeah, that was a mistake. I made a commit to address it.

There's one last thing I want to verify, so don't feel rushed to review/verify this PR yet. It's something I can check out when I get time. The diff below came from reverting the commit that introduced the change that caused a long string to be passed to `process-send-string`, but I'm not confident we actually want to revert this part.

```
  (let* ((append-string ; strip setup code from Elpy
          (if (string-match "import sys, codecs, os, ast;__pyfile = codecs.open.*$" string)
  (let* ((append-string ; strip setup code from Python shell
          (if (string-match "import codecs, os.*__pyfile = codecs.open.*$" string)
```

elpy-shell.el
```diff
@@ -146,6 +146,25 @@ the code cell beginnings defined here."
 (defvar elpy--shell-last-py-buffer nil
   "Help keep track of python buffer when changing to pyshell.")
 
+(defvar elpy-shell--python-version-cache nil)
+(defvar elpy-shell--python-interpreter-cache nil)
```
eihli: Maybe this is poorly named. The var just keeps track of the value of `python-shell-interpreter` from the last time `elpy-shell-get-python-version` was run so that if the value of `python-shell-interpreter` has changed then we can make a call to get the new python version. So it's not really a "cache" in the sense of a performance improvement by storing some value we don't expect to change often.

elpy-shell.el
```diff
@@ -146,6 +146,25 @@ the code cell beginnings defined here."
...
+
+(defun elpy-shell-get-python-version ()
+  "Return the python interpreter version"
+  (if (or
+	   (not (eq elpy-shell--python-interpreter-cache
+				python-shell-interpreter))
+	   (null elpy-shell--python-version-cache))
```
eihli: These lines check to see if we need to make the expensive call to python --version. It assumes the only way the version can change is if python-shell-interpreter changes. Is that a safe assumption? It works for my workflow. But is there a tool, like pyenv, where the python-shell-interpreter is some wrapper script that calls out to different versions based on env vars or dotfiles or something.

elpy-shell.el
galaunay: I am personally okay with it, as long as the docstring of the variable explains what it is properly.

Or maybe `elpy-shell--python-version-cached-interpreter' ?

elpy-shell.el
galaunay: If `python-shell-interpreter` is just `"python"`, the python version will depend on the first python binaries found in the PATH (if you use virtualenv for exemple, it can change).
But this can be addressed by storing the full python path instead of just the name : 
``` elisp
 (if (or
	   (not (eq elpy-shell--python-interpreter-cache
				(executable-find python-shell-interpreter)))
	   (null elpy-shell--python-version-cache))
```

The problem with pyvenv is more complicated to solve as even the full path will be the same...
I don't have any solution at the moment.

eihli:
I think I learned of a much better solution.

https://debbugs.gnu.org/cgi/bugreport.cgi?bug=12440

> > > in Emacs 24.2, the following code
> > >
> > >   (let* ((process-connection-type t)
> > >          (process (start-process "cat" (current-buffer) "cat")))
> > >     (process-send-string process (format "%s\n" (make-string 6000 ?a))))
> > >
> > > inserts only 4095 characters in the current buffer. Setting
> > > process-connection-type to nil will insert all characters.
> > 
> > I can confirm that this bug is still present in Emacs 27.
> > 
> > It's odd that this doesn't cause more problems than it does in practice,
> > but I guess it's rare that we talk to processes this way.
> 
> IMO, when a Lisp program intends to communicate with a subprocess this
> way, it should indeed set process-connect-type to nil.  It makes no
> sense to me to use PTYs to communicate massive amounts of data in
> batch-like manner.

Yesterday, I manually verified this solved my issue. Today, I upgraded to MacOS Catalina and I'm unable to even reproduce the issue on master. :( That makes it difficult to write a test since I can't verify the test correctly fails without the fix. Along with that, I'm quite unfamiliar with Emacs-lisp in general and I'm not quite sure how the test should be written.

galaunay:
> https://debbugs.gnu.org/cgi/bugreport.cgi?bug=12440

Nice, very useful !

I found this in the manual regarding `process-connection-type`:
```
An asynchronous process is controlled either via a “pty”
(pseudo-terminal) or a “pipe”.  The choice of pty or pipe is made when
creating the process, by default based on the value of the variable
‘process-connection-type’ (see below).  If available, ptys are usually
preferable for processes visible to the user, as in Shell mode, because
they allow for job control (‘C-c’, ‘C-z’, etc.) between the process and
its children, and because interactive programs treat ptys as terminal
devices, whereas pipes don’t support these features.  However, for
subprocesses used by Lisp programs for internal purposes (i.e., no user
interaction with the subprocess is required), where significant amounts
of data need to be exchanged between the subprocess and the Lisp
program, it is often better to use a pipe, because pipes are more
efficient. 
```

So IMHO, the best way to go for python shell is to keep using ptys.
However, we could make Elpy uses pipes for macOS as a temporary workaround until the related bug get fixed in Emacs ?

> That makes it difficult to write a test since I can't verify the test correctly fails without the fix. Along with that, I'm quite unfamiliar with Emacs-lisp in general and I'm not quite sure how the test should be written.

Your test looks fine. Given the situation, I don't think we can do better than just testing that it works.

galaunay:
I took the liberty to make the few necessary adjustments.

Thanks for this !

Issue #1430, opened 2018-08-05
IPython Shell magic function %hist outputs lots of junk history 


shelper:
# Summary

I am using spacemacs with Emacs V26.1, with package elpy, on a Macbook. I believe this also happens on Windows. the issue is when I run the magic function %hist in the IPython shell, i got lots of information like below:

```
import codecs, os;__pyfile = codecs.open('''/var/folders/y1/197_jjzs1yv2qz0m26z1jkxw0000gn/T/pysfAkcq''', encoding='''utf-8''');__code = __pyfile.read().encode('''utf-8''');__pyfile.close();os.remove('''/var/folders/y1/197_jjzs1yv2qz0m26z1jkxw0000gn/T/pysfAkcq''');exec(compile(__code, '''/var/folders/y1/197_jjzs1yv2qz0m26z1jkxw0000gn/T/pysfAkcq''', 'exec'));
import codecs, os;__pyfile = codecs.open('''/var/folders/y1/197_jjzs1yv2qz0m26z1jkxw0000gn/T/pythwlpb''', encoding='''utf-8''');__code = __pyfile.read().encode('''utf-8''');__pyfile.close();os.remove('''/var/folders/y1/197_jjzs1yv2qz0m26z1jkxw0000gn/T/pythwlpb''');exec(compile(__code, '''/var/folders/y1/197_jjzs1yv2qz0m26z1jkxw0000gn/T/pythwlpb''', 'exec'));
import codecs, os;__pyfile = codecs.open('''/var/folders/y1/197_jjzs1yv2qz0m26z1jkxw0000gn/T/pyUwYsfX''', encoding='''utf-8''');__code = __pyfile.read().encode('''utf-8''');__pyfile.close();os.remove('''/var/folders/y1/197_jjzs1yv2qz0m26z1jkxw0000gn/T/pyUwYsfX''');exec(compile(__code, '''/var/folders/y1/197_jjzs1yv2qz0m26z1jkxw0000gn/T/pyUwYsfX''', 'exec'));
```
anyone had the same issue and is there a way to remove these?

# Steps to reproduce
run `run-python`, then run couple of command, e.g., `print('helloworld') and run `%hist`

# My configuration
## OS
mac os 10.13

## Result of `(elpy-config)`
```
Virtualenv........: None
RPC Python........: 3.5.2 (/Users/zyuan/anaconda/bin/python)
Interactive Python: ipython (/Users/zyuan/anaconda/bin/ipython)
[...]

PR #1684, opened 2019-10-03
Add string sent by Elpy to the comint history

galaunay:
# PR Summary
Follow #1430.

When sending a statement/region/defun/group/codecell, Elpy is now adding the sent string to the comint history. 
This allows to get those inputs back from the shell with `comint-previous-input` and similar.

# PR checklist
- [x] Add an option because this could be quite annoying if enabled by default.

### DIFF
--- a/docs/ide.rst
+++ b/docs/ide.rst
@@ -403,6 +403,12 @@ accidental code execution, e.g.:
 
    Also bound to :kbd:`C-c C-c`.
 
+.. option:: elpy-shell-add-to-shell-history
+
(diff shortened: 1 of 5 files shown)


rgemulla:
I think this could be useful in some cases and annoying in others. If added, I think this feature should be optional. Perhaps an option that controls what is kept in the history (e.g., only python-shell-send-statement).

galaunay:
I agree, that's why it is in pause at the moment.
I wanted to find a solution that wouldn't involve an additional option, but I think this is the only way to go.

PR #1686, opened 2019-10-08
fix #1685

### LINKED ISSUE #1685: elpy-goto-definition not working after elpy update
# Summary

Hi, I was running elpy with versión 20190130.2109 without any issues relating to this - I could view definitions originating from other files and within same files. 
Since updating (I noticed there is a new rpc-venv folder on disk now) it doesn't work. 
I cannot even view definitions within same file. 

I have tried to verify jedi works, using some hints given to another user here (Chinese) and my jedi seems to work - at least with the python exe I intend to use. 

# Steps to reproduce

elpy-goto-definition 

# My configuration
## OS

Windows 10

## Result of `(elpy-config)`
```
Emacs.............: 26.1

Elpy..............: 1.31.0

Virtualenv........: py36 (c:/PF/Anaconda3/envs/py36)

Interactive Python: C:/Program Files/Anaconda3/envs/py36/python.exe . (C:/Program Files/Anaconda3/envs/py36/python.exe)

RPC virtualenv....: rpc-venv (c:/Users/<user>/AppData/Roaming/.emacs.d/elpy/rpc-venv)

Python...........: c:/PF/Anaconda3/envs/py36/pythonw.exe 3.6.9 (c:/PF/Anaconda3/envs/py36/pythonw.exe)

Jedi.............: 0.15.1

Rope.............: Not found

Autopep8.........: 1.4.4

Yapf.............: 0.28.0

Black............: 19.3b0

Syntax checker....: flake8.exe (c:/PF/Anaconda3/envs/py36/Scripts/flake8.exe)

 

The python interactive interpreter (C:/Program

Files/Anaconda3/envs/py36/python.exe) is not installed on the current virtualenv

(c:/PF/Anaconda3/envs/py36/). The system binary (C:/Program

Files/Anaconda3/envs/py36/python.exe) will be used instead.

```
## Elpy configuration in my init.el
```
(pyvenv-activate "C:/PF/Anaconda3/envs/py36")

(load "elpy")

(elpy-enable)

(require 'python-mode)

(autoload 'python-mode "python-mode" "Python Mode." t)

(add-to-list 'auto-mode-alist '("\\.py\\'" . python-mode))

(add-to-list 'interpreter-mode-alist '("python" . python-mode))

(when (require 'flycheck nil t)

  (setq elpy-modules (delq 'elpy-module-flymake elpy-modules))

  (add-hook 'elpy-mode-hook 'flycheck-mode))

 

(global-eldoc-mode -1)

 

(define-key global-map (kbd "C-M-g") 'elpy-goto-definition)

(setq python-shell-interpreter "C:/Program Files/Anaconda3/envs/py36/python.exe")
```

kevindragon:
# PR Summary

In Jedi, function [jedi.create_environment](https://github.com/davidhalter/jedi/blob/a6fcf779d41862de50d25cbe964c64831c7d64bc/jedi/api/environment.py#L344) accept a python executable path (windows, linux
and macOS) or a directory path which containing the bin directory. But on
windows the variable `deactivated-environment` has got incorrect path, so elpy.jedibackend.JediBackend can not be initialized with an error 'C:\xxxx\Scripts\python.exe missing'

BTW, may be we need a right way to handle the difference between different OS.

# PR checklist

## For new features only:

### DIFF
--- a/elpy-rpc.el
+++ b/elpy-rpc.el
@@ -250,6 +250,8 @@ BODY."
                                                pyvenv-post-deactivate-hooks))
            (deactivated-environment
             (or pyvenv-virtual-env
+                (when (string-equal system-type "windows-nt")
+                  (executable-find elpy-rpc-python-command))
                 ;; global env
                 (directory-file-name
                  (file-name-directory


galaunay:
Hi, 
Thanks for the PR.

The `deactivated-environment` should be set to the virtualenv path (the folder containing `/bin/python.exe`), not directly the python binaries.
I am actually surprised setting `deactivated-environment` to `(executable-find elpy-rpc-python-command)` works.

Can I ask what is the result of `(executable-find elpy-rpc-python-command)` on windows ?

kevindragon:
> Hi,
> Thanks for the PR.
> 
> The `deactivated-environment` should be set to the virtualenv path (the folder containing `/bin/python.exe`), not directly the python binaries.
> I am actually surprised setting `deactivated-environment` to `(executable-find elpy-rpc-python-command)` works.
> 
> Can I ask what is the result of `(executable-find elpy-rpc-python-command)` on windows ?

the result of `(executable-find elpy-rpc-python-command)` is `c:/Users/jiangkx/AppData/Local/Continuum/anaconda3/pythonw.exe`

original value of `deactivated-environment` on windows is `c:/Users/jiangkx/AppData/Local/Continuum`,  function jedi.create_environment can not find `Script/python.exe`

may be, I should change the variable `environment` passed to jedi, other than change `deactivated-environment` directly

galaunay:
The problem is quite complex and linked to the fact jedi does not seem to be able to work on anaconda system environments (see description on PR 1688).

It is not easily fixable, but I make PR #1688 to allow Anaconda users to get Elpy to work with the option `(setq elpy-rpc-virtualenv-path 'current)`.

Could you try PR #1688 with the option  `(setq elpy-rpc-virtualenv-path 'current)` and tell me if it works for you ?

galaunay:
I found a better solution and merged it today. Could you try the last version of Elpy and tell me if the issue still exists ?

kevindragon:
After upgrade, I still got: `elpy-xref--definitions: No definition found`

galaunay:
Elpy should now use the full binaries path instead of the environment path (as you mentioned in your previous post).

Just to check it is properly working, could you post the result of `elpy-config` ?
And the result of `(executable-find elpy-rpc-python-command)` and `(elpy-rpc-get-or-create-virtualenv)` when in a python buffer ?

kevindragon:
It's working now, thanks @galaunay , I think this pr should be closed.

galaunay:
Thanks for the feedback.
I am closing it.

Issue #1695, opened 2019-10-13
Failed test: elpy-rpc-get-virtualenv-path-should-return-current-venv-path


memeplex:
I've forked the project and followed the instructions in `CONTRIBUTING.md`. I just want to let you know that one unit test is not passing in my environment:

```
Ran 414 tests in 99.380 seconds
1 unexpected results:
   FAILED  elpy-rpc-get-virtualenv-path-should-return-current-venv-path
```

galaunay:
The new branch tests seem to pass.
Please re-open if it still relevant;

PR #1696, opened 2019-10-13
Redefine compile/debug commands to more sensible defaults (Fix #1693)

### LINKED ISSUE #1693: Redefine compile-command and gud-pdb-command-name to more sensible value
The defaults:

```
compile-command = make -k
gud-pdb-command-name = pdb
```

doesn't make much sense.

Why not redefine them, in a very conservative fashion anyway:

```
compile-command = python
gud-pdb-command-name = python -m pdb
```

?

memeplex:
This fixes issue #1693 

Tests are passing except for https://github.com/jorgenschaefer/elpy/issues/1695 which is totally unrelated to this change.

### DIFF
--- a/docs/ide.rst
+++ b/docs/ide.rst
@@ -657,9 +657,9 @@ advantages this brings.
 .. command:: elpy-set-test-runner
 
    This changes the current test runner. Elpy supports the standard
-   unittest discovery runner, the Django discovery runner, nose and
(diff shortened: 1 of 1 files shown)


memeplex:
A minor remark, for what it's worth: you could use `setq-local x` instead of the more verbose `set (make-local-variable 'x)`.

Issue #1531, opened 2019-02-11
Maintainership of python.el


memeplex:
@galaunay @jorgenschaefer do you know who is currently maintaining python.el? Can you commit changes there?

I'm asking here because I've submitted some patches to the emacs tracker months ago that fix problems with inferior ipython and received little to no answer. Besides, there is [no commit](http://git.savannah.gnu.org/cgit/emacs.git/log/lisp/progmodes/python.el?h=emacs-26) since about a year, except for a copyright update. Development seems pretty stalled and that's starting to worry me. I even offered maintaining it but nobody answered.

I understand it's not elpy fault but since python.el is the foundation of elpy I'm asking you some kind of advice about how to proceed.

galaunay:
I think you were looking at the emacs 26 branch.
The [master branch](http://git.savannah.gnu.org/cgit/emacs.git/log/lisp/progmodes/python.el) still show some activity on `python.el` (not much, I agree, in the past year).

The current [MAINTAINERS](http://git.savannah.gnu.org/cgit/emacs.git/tree/admin/MAINTAINERS) file doesn't mention anybody being responsible for python.el.

memeplex:
Ah ok, you're right, at least is not completely dead...

memeplex:
I've been thinking about this situation. I believe the main problem is that python.el was subsumed by emacs. It's understandable for emacs to ship basic major modes for popular languages, but python.el is far from basic. Now, emacs is a large piece of software and contributing to it is not as easy as if it were hosted here: legal issues, a peculiar bug tracker, etc. (don't take me wrong, I contribute to emacs and it's not rocket science, it's just a harder proposition -vs github/gitlab- for the average potential contributor these days). More importantly, maintainers have their priorities and it's understandable that python is not always at the top of their list.

Nowadays ipython (+prompt_toolkit, +jedi) is offering a better cross-platform REPL experience than emacs. It's not difficult to improve python inferior mode to provide competitive features, I've even suggested simple advices to keep history colorized. But more basic things are broken, in particular ipython integration has been always difficult to keep working straight because of the faster release cycle and dynamism of ipython vs python.el. Some of these features are important to elpy as a python IDE and the decision to rely on python.el as included in the latest stable emacs release has its costs.

Here are some options I've thought of:

[...]

PR #1698, opened 2019-10-13
Use native python fontification when sending input to shell (fix #1428)

### LINKED ISSUE #1428: Full inferior mode input font locking
I know this is a comint issue and I've already reported at https://debbugs.gnu.org/cgi/bugreport.cgi?bug=32344 (please see details there).

But maybe you're interested in adding this workaround to elpy (perhaps optionally depending on a customization option) in order to get full input font locking, it looks really nice!

```elisp
(advice-add 'comint-send-input
            :around (lambda (f &rest args)
                      (when (eq major-mode 'inferior-python-mode)
                        (cl-letf (((symbol-function 'g)
                                   (symbol-function 'add-text-properties))
                                  ((symbol-function 'add-text-properties)
                                   (lambda (start end properties &optional object)
                                     (unless (eq (nth 3 properties) 'comint-highlight-input)
                                       (g start end properties object)))))
                          (apply f args)))))
```

memeplex:
# PR Summary

Use full python font locking in shell, even when sending input from an elpy buffer. Fixes #1428 

**Warning**: this is not ready for merging because it requires some changes upstream that I'm pushing here: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=32344

Specifically, these two very simple patches should be merged:

* https://debbugs.gnu.org/cgi/bugreport.cgi?msg=20;att=1;bug=32344;filename=0001-Option-for-comint-to-honour-original-input-highlight.patch

* https://debbugs.gnu.org/cgi/bugreport.cgi?msg=35;att=1;bug=32344;filename=0001-Keep-python-shell-input-colorization-in-comint.patch

Alternatively we could add an "experimental module" (disabled by default) to get this [1] working now using some elisp hack:

```elisp
(advice-add 'elpy-shell--insert-and-font-lock
            :around (lambda (f string face &optional no-font-lock)
                      (if (not (eq face 'comint-highlight-input))
                          (funcall f string face no-font-lock)
                        (funcall f string face t)
                        (python-shell-font-lock-post-command-hook))))
```
But let's hope my patches get merged upstream.

# PR checklist

---

[1] And probably other stuff that require monkey-patching hacks. Related to https://github.com/jorgenschaefer/elpy/issues/1531

### DIFF
--- a/elpy-shell.el
+++ b/elpy-shell.el
@@ -519,8 +519,10 @@ complete). Otherwise, does nothing."
   (let ((from-point (point)))
     (insert string)
     (if (not no-font-lock)
-        (add-text-properties from-point (point)
-                             (list 'front-sticky t 'font-lock-face face)))))
+        (if (eq face 'python-font-lock)
+            (python-shell-font-lock-post-command-hook)
+          (add-text-properties from-point (point)
+                               (list 'front-sticky t 'font-lock-face face))))))
 
 (defun elpy-shell--append-to-shell-output (string &optional no-font-lock prepend-cont-prompt)
   "Append the given STRING to the output of the Python shell buffer.
@@ -540,7 +542,7 @@ Prepends a continuation promt if PREPEND-CONT-PROMPT is set."
                    (lines (split-string string "\n")))
               (goto-char mark-point)
               (elpy-shell--insert-and-font-lock
-               (car lines) 'comint-highlight-input no-font-lock)
+               (car lines) 'python-font-lock no-font-lock)
               (when (cdr lines)
                   ;; no additional newline at end for multiline
                   (dolist (line (cdr lines))
@@ -548,11 +550,11 @@ Prepends a continuation promt if PREPEND-CONT-PROMPT is set."
                     (elpy-shell--insert-and-font-lock
                      prompt 'comint-highlight-prompt no-font-lock)
                     (elpy-shell--insert-and-font-lock
-                     line 'comint-highlight-input no-font-lock)))
+                     line 'python-font-lock no-font-lock)))
                 ;; but put one for single line
                 (insert "\n"))
           (elpy-shell--insert-and-font-lock
-           string 'comint-highlight-input no-font-lock))
+           string 'python-font-lock no-font-lock))
         (set-marker (process-mark process) (point)))))))
 
 (defun elpy-shell--string-head-lines (string n)


galaunay:
Nice,
I have this hack in my init.el for so long that I forgot it was a hack.
Any idea of the timescale for you patches to be merged ?

memeplex:
> Any idea of the timescale for you patches to be merged ?

Not at all. Maybe @npostavs could make a more informed guess here.

rgemulla:
I like this idea! I did a quick try & it did not work in my setup, however. In particular, it seems to break when continuation lines are used (```elpy-shell-echo-input-cont-prompt``` set to t). When pasting multiline statements, the python font lock hook seems to get confused with the continuation prompt. 

One way to fix this may be to insert the code fragment into a temporary buffer, propertize it there using the python font lock, and then copying it line by line to the shell buffer.

memeplex:
Have you installed the corresponding comint and python.el patches that should be merged into emacs 27 for this to work? You also have a more hacky version in the wiki using advices.

rgemulla:
Yes, it's the call to the python font lock hook that breaks.

memeplex:
> One way to fix this may be to insert the code fragment into a temporary buffer, propertize it there using the python font lock, and then copying it line by line to the shell buffer.

AFAICR this is exactly what is done.

The changes here just disable recolorizing the already colorized text. If there is something wrong in the way text is colorized in some corner cases, I believe the problem should be addressed separately.

rgemulla:
I apologize, the problem arise because I already applied PR #1710. I now changed the PR slightly so that the advice-based hack from the doc still works. It's not nice, but at least it does not break.

That being said, if continuation prompts are turned on and the hack are being used, then the font lock of the continuation prompt is lost (at least in my setup). 

Also, whether or not continuation prompts are used, the font lock in the shell buffer does not seem to match the one in the Python buffer (e.g., integers and variable assignments are not font-locked in the shell buffer). Is this the same with this PR?

One idea that I had when originally writing this code was to keep track of the text properties from the python buffer throughout. I never got to implement this. The advantage would be that the font lock in the shell buffer exactly matches the python buffer. It should also be faster since there is no separate font lock.

rgemulla:
Actually, the only thing that is fontified differently in the python buffer than in the shell buffer seems to be numbers.

rgemulla:
The problem with not fontifying numbers arises also directly in the shell buffer (i.e., when entering numbers there); not sure why.

rgemulla:
I figured out the reason for the difference in number highlighting: I have highlight-numbers-mode enabled python-mode, and it's not automatically enabled in inferior-python-mode as well. Manually enabling it solves this issue.

The only remaining problem is that the fontification of the continuation prompt is destroyed.

rgemulla:
Ok, found the issue, fix below. I suggest to change this PR such that only ```elpy-shell--insert-and-font-lock``` is changed (but not ```elpy-shell--append-to-shell-output```). A definition that respects continuation prompts and keeps fontification of inputs is:

```elisp
(defun elpy-shell--insert-and-font-lock (string face &optional no-font-lock)
  "Inject STRING into the Python shell buffer."
  (let ((from-point (point)))
    (insert string)
    (unless no-font-lock
      (if (eq face 'comint-highlight-input)
          (cl-letf* ((stp (symbol-function 'set-text-properties))
                     ((symbol-function 'set-text-properties)
                      (lambda (start end plist &optional object)
                        (while (< start end)
                          (unless (eq (get-text-property start 'face) 'comint-highlight-prompt)
                            (funcall stp start (+ 1 start) plist object))
                          (setq start (+ 1 start))))))
            (python-shell-font-lock-post-command-hook))
        (add-text-properties
         from-point (point)
         (list 'front-sticky t 'font-lock-face face))))))
```

Only the advice for comint-input-prompt is then needed (until addressed upstream).

rgemulla:
This fix improves things, but it is still not perfect. Since ```python-shell-font-lock-post-command-hook``` does not know that continuation prompts should be ignored, the only way to improve it further is to replace ```python-shell-font-lock-post-command-hook``` by a custom function, I think.

rgemulla:
I found a simpler and better solution, see PR #1711. Let me know what you think.

galaunay:
#1711 works fine for me.

Will we have some compatibility issue if we merge it and the patches to Emacs get merged as well ?

rgemulla:
No, #1711 solely fontifies the source code before it's being echoed in the shell buffer. In fact, ```my-comint-highlight-input-advice``` mentioned above is currently still necessary.

@memeplex What are your thoughts?

Issue #1709, opened 2019-10-29
Unexpected EOF error while evaluating objects inline


bibsian:
# Summary
I get an unexpected EOF error while evaluating objects in line.

# Steps to reproduce
### Inline evaluation error
``` python
y = [1, 2, 3, 4]
```
If I create a list like above, highlight a number within the list, and evaluate the integer with `C+c C+c`, I get the following error:
``` python
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/bibsian/anaconda3/envs/fastai/lib/python3.6/ast.py", line 35, in parse
    return compile(source, filename, mode, PyCF_ONLY_AST)
  File "/home/bibsian/Desktop/lvdata/lvdata/scipy/nb.py", line 13
    Y = [1
         ^
SyntaxError: unexpected EOF while parsing
```
# My configuration
## OS
``` bash
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.3 LTS
Release:	18.04
Codename:	bionic
```

## Result of `(elpy-config)`
``` elpy
Elpy Configuration

Virtualenv........: fastai (/home/bibsian/anaconda3/envs/fastai)
RPC Python........: 3.6.9 (/home/bibsian/anaconda3/envs/fastai/bin/python)
Interactive Python: python (/home/bibsian/anaconda3/envs/fastai/bin/python)
Emacs.............: 26.3
Elpy..............: 1.31.0
Jedi..............: 0.15.1
Rope..............: Not found (0.14.0 available)
Autopep8..........: Not found (1.4.4 available)
Yapf..............: Not found (0.28.0 available)
Black.............: Not found (19.10b0 available)
Syntax checker....: flake8 (/home/bibsian/anaconda3/envs/fastai/bin/flake8)

```

## Elpy configuration in my init.el
``` elpy
;;Melpa repo

;; Added by Package.el.  This must come before configurations of
;; installed packages.  Don't delete this line.  If you don't want it,
;; just comment it out by adding a semicolon to the start of the line.
;; You may delete these explanatory comments.
(package-initialize)

(require 'package)
(add-to-list 'package-archives
             '("melpa-stable" . "https://stable.melpa.org/packages/"))

(setq ring-bell-function 'ignore)

;;;;;;;;;;;;;;;;;;;;;;;;;;;
;;; Installed pacakges ;;;;
;;;;;;;;;;;;;;;;;;;;;;;;;;;

[...]

PR #1710, opened 2019-11-01
Bugfix for sending regions to shell

rgemulla:
This fixes two problems when sending regions to the python interpreter via the `elpy-shell-send-region` functions.

Consider input
```
1 ; 2 ; 3
```
Mark only ```2``` and call `elpy-shell-send-region`.

Before patch:
- ```1; 2``` was sent to the interpreter.
- ```Sent: 1; 2; 3``` was shown in the echo area.

After patch:
- ```2``` is sent to the interpreter
- ```Sent: 2``` is shown in the echo area.

I documented details and fix of this problem in the PR's code. The patch also works correctly when multiline statements are selected.

### DIFF
--- a/elpy-shell.el
+++ b/elpy-shell.el
@@ -562,8 +562,13 @@ Prepends a continuation promt if PREPEND-CONT-PROMPT is set."
                   ;; no additional newline at end for multiline
                   (dolist (line (cdr lines))
                     (insert "\n")
-                    (elpy-shell--insert-and-font-lock
-                     prompt 'comint-highlight-prompt no-font-lock)
+                    (let ((from-point (point)))
+                      (elpy-shell--insert-and-font-lock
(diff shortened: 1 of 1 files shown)


rgemulla:
It's a small PR and ready to be merged in my point of view. I do not merge it right away since I haven't been active for a while and I feel that @galaunay should have a look first.

rgemulla:
I added another small commit related to this PR. When a continuation prompt is printed in the shell, it's now propertized as "output" field just like the actal prompt.

galaunay:
Seems perfect to me.
It fixes issue #1709 as well.

rgemulla:
Thanks for the quick feedback!

galaunay:
Thanks for fixing this :).
I Just added some tests to be sure something won't break if `python-shell-buffer-substring` changes.

PR #1741, opened 2020-01-09
Make after-change-function save match data.

kenmanheimer:
# PR Summary
elpy-folding--mark-foldable-lines was disrupting editing activities, like query replace, because it failed to save match data. This PR fixes that, wrapping the effective body of the function in save-match-data. I also wrapped resulting long lines.

# PR checklist

## For new features only:


galaunay:
Legitimate, thanks for the PR.

I am happy to see that someone is using the folding module.
It has been implemented not so long ago, so feel free to reach us if it is misbehaving (again) or if you have ideas to make it better.

PR #1770, opened 2020-04-10
Fix typo in elpy-shell-add-to-shell-history docstring & enhance flow

sten0:
Post-edits, this docstring makes it sound like this toggle enables a
kill-ring or clipboard history for the Python shell, which sounds like
it's always a good thing.  That said, it's disabled by default.  I
believe it would be useful to let the user know why it's disabled by
default.

Also, I think it would be useful to have a follow up commit that
explains how it "affects the following functions:…".

# PR checklist

## For new features only:

### DIFF
--- a/elpy-shell.el
+++ b/elpy-shell.el
@@ -155,10 +155,11 @@ the code cell beginnings defined here."
   :group 'elpy)
 
 (defcustom elpy-shell-add-to-shell-history nil
-  "If Elpy should make the code sent to the shell available in the
-shell history. This allows to use `comint-previous-input' in the
-python shell to get back the pieces of code sent by Elpy. This affects
-the following functions:
+  "Toggle that affects whether Elpy makes the code that is sent
+to the Python shell available in the shell's history.  If
+enabled, use `comint-previous-input' in the Python shell to get
+back the pieces of code sent by Elpy. This affects the following
+functions:
 - `elpy-shell-send-statement'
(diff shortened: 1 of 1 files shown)


gopar:
Has there been other ppl confused/misunderstood about that docstring?

Otherwise LGTM.

sten0:
Hi @gopar, thank you for reviewing!  By the way, do you know why `elpy-shell-add-to-shell-history` is disabled by default?  If you or @galaunay aren't planning to document that in a follow-up commit, then I'd like to include it in the PR for completeness sake.  If necessary I'd also be happy to investigate how it changes the behaviour of the listed functions, and document this for the same reason.

gopar:
@sten0 I'm not sure why it's disabled by default. Must of been a decision from a while ago.

Yeah, if you can document that extra bit, that would be awesome :) 

Ping me whenever it's ready. No rush. Thanks! \o/

galaunay:
This was disabled by default because we suspected it could be annoying (e.g. when you send big chunks of code that you don't want to find back in your history).
That's also why it only affects some functions (`send-defun`, `send-defclass` and `send-buffer` were set aside).

I agree the docstring should at least mention that.
Or if you have a another solution it could be the opportunity to make it better ?

gopar:
@sten0 let me know if you wanted to do that follow up doc change. otherwise i'll commit this.

Issue #1774, opened 2020-04-14
hiding docstrings


drdv:
`elpy-info-docstring-p` doesn't handle functions whose signature is on several lines, e.g.,

```
def test(argument1,
         argument2):
    """Test function."""
```
The function `python-info-docstring-p` (in `python.el`) doesn't have this problem. Is there any reason not to use it directly?

galaunay:
I guess there was one at some point.

But you are right that it works better with `python-info-docstring-p`.
I made a PR to make the switch.

drdv:
Thanks. I am closing this one. I left a question at #1776

PR #1776, opened 2020-04-18
Better docstring detection when folding

galaunay:
# PR Summary
Follow #1774.

In the folding module, we use `elpy-info-docstring-p` to check if the cursor in on a docstring.
`python-info-docstring-p` does a way better job at that, and there is no reason (anymore ?) not to use it.

# PR checklist


drdv:
Thanks, this is a nice change and it permits cleaning up other things as well. I would like to suggest another modification related to hiding docstrings. I am not, however, sure what would be the most convenient way for making a PR that depends on this PR.

gopar:
LGTM 

Sorry I haven't active :(

Issue #1773, opened 2020-04-13
Jedi type hint support


cc-a:
Support for type hints should be included in the next release of Jedi. I've started work on making use of this feature in elpy (see [my fork](https://github.com/cc-a/elpy/commit/7a6e27222c3bebdbaeb83c63676e99b0d93df133). Please let me know if you'd be interested in a PR for this functionality.

galaunay:
Very much indeed, thanks.
I'll gladly accept this PR.

PR #1785, opened 2020-05-02
Completion type hints

cc-a:
# PR Summary

See #1773

This adds support for use of the `get_type_hint` method for Jedi completions that was added 0.17.0. For appropriate proposals type hints are added as part of the completion annotation.

Please let me know if there are any other locations where `get_type_hint` could be useful and I'll happily take a look.

For completion candidates of functions the type hints can be quite long so it might be sensible to introduce a truncation. Let me know any thoughts.

# PR checklist

## For new features only:

### DIFF
--- a/elpy/jedibackend.py
+++ b/elpy/jedibackend.py
@@ -46,7 +46,7 @@ class JediBackend(object):
                                 for proposal in proposals)
         return [{'name': proposal.name.rstrip("="),
                  'suffix': proposal.complete.rstrip("="),
-                 'annotation': proposal.type,
+                 'annotation': get_annotation(proposal),
                  'meta': proposal.description}
                 for proposal in proposals]
 
@@ -263,7 +263,13 @@ class JediBackend(object):
             else:
                 onelinedoc = onelinedoc[0]
             if onelinedoc == '':
-                onelinedoc = "No documentation"
+                if definition.type == "instance":
+                    try:
+                        onelinedoc = definition.get_type_hint()
+                    except AttributeError:
+                        onelinedoc = "No documentation"
+                else:
+                    onelinedoc = "No documentation"
             return {"name": name,
                     "doc": onelinedoc}
         return None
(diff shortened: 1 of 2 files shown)


galaunay:
It looks nice.

There is some tests that are failing (travis doesn't show up in PRs anymore, but here is the [link](https://travis-ci.org/github/jorgenschaefer/elpy/builds/682624223)).

We are still supporting python2.7 so it would be nice to make your additional test conditional.
There is another error linked to `FileNotFoundError`, but I am not sure where it is coming from...

cc-a:
Resolved the existing errors but there is a random smattering of failures. I can't see how any of them would be related to my changes and I've not been able to recreate them locally or in Docker containers. I'm open to suggestions on how to proceed.

galaunay:
Unfortunately, those random failures are a long-standing problem I cannot really get my head around... I restarted the tests and your commit looks fine.

It works nicely, but it can take quite some space and become unreadable depending on the typing. For examples, for the json package on a clean Emacs, this is what I get:
img-2020-05-27-172105

Some ideas that can help reduce the clutter:
- It would be nice to remove the spaces and newlines (with `hint = re.sub('\s+', ' ', hint)` or similar).
- Maybe we don't need the `statement` and `function` annotations if we already have the type hint (should be pretty explicit ?).
- Having the class `__init__` arguments would be nice as well, but I don't know if it is doable...

Quickly trying to address the first two points, this is what I get:
``` python
def get_annotation(proposal):
    if proposal.type in ("instance", "statement", "param", "function"):
        try:
            hint = proposal.get_type_hint().replace('\n', ' ')
            hint = re.sub("\\s+", " ", hint)
            hint = re.sub("\\s*= \\.\\.\\.\\s*", " ", hint)
            hint = re.sub("\\s+,", ",", hint)
            hint = "  {}".format(hint)
        except (AttributeError, TypeError, NotImplementedError, NameError):
            hint = ""
        if hint == "":
            return proposal.type
        else:
            return hint
    else:
        return proposal.type
```
img-2020-05-27-175011

A bit better but still quite long. In my opinion we will need to make this feature optional (some users may not want this additional information) and maybe strip the end of the annotations to keep the company window to a decent size.

Hope it makes sens.

galaunay:
Another solution would be to put the type hints as "meta" instead of "annotation".
This will display the type hints in the minibuffer (where there is more space).
It sounds like a good solution to me: it is not cluttering the company menu and still make the information accessible.

PR #1787, opened 2020-05-05
Teach elpy how to goto-definition in async code

danking:
# PR Summary

I haven't tested this from within Emacs as I'm not really sure how to use changes I've made to the elpy python code. Nonetheless, I'm running into issues with asyncio code. In the code below, I'm unable to go to the definition of `y`.  I'm fairly certain the issue is resolved in Jedi 0.17.0, but only if one uses `Script.goto`, which now accepts the line and column arguments. It's deprecated to pass such arguments to the `Script` directly.

```
async def f(x, y):
    return x, y


async def foo():
    x, y = await f(1, 3)

    x

    y
```

If I stick the above in a script and run it I get exactly what I want when using `goto`:

```
In [5]: b = jedi.create_environment('/usr/local/')                                                                                                                                                                                                   
In [6]: script = jedi.Script(source=""" 
   ...: async def f(x, y): 
   ...:     return x, y 
   ...:  
   ...:  
   ...: async def foo(): 
   ...:     x, y = await f(1, 3) 
   ...:  
   ...:     x 
   ...:  
   ...:     y 
   ...: """, line=10, column=4, path="/foo.py", encoding='utf-8', environment=b)                                                                                                                                                                                                         
In [7]: script.goto(line=9, column=4)                                                                                                                                                                                                                
Out[7]: [<Name full_name='foo.foo.x', description='x, y = await f(1, 3)'>]

```

# PR checklist

### DIFF
--- a/elpy/jedibackend.py
+++ b/elpy/jedibackend.py
@@ -366,6 +366,8 @@ def run_with_debug(jedi, name, *args, **kwargs):
     re_raise = kwargs.pop('re_raise', ())
     try:
         script = jedi.Script(*args, **kwargs)
+        if name == 'goto_definitions':
+            return script.goto(line=kwargs['line'], column=kwargs['column'])
         return getattr(script, name)()
     except Exception as e:
         if isinstance(e, re_raise):


danking:
If you cannot accept this PR, could you let me know how to make changes from elpy/elpy/jedibackend.py be used in Emacs? AFAICT, my changes not being used. I'd love to get definition jumping working in async code. I also cannot find an elpy RPC log that would help me better understand whether my changes are working as expected.

danking:
ah, I found elpy-rpc-restart. disregard my comment above.

danking:
OK, can confirm, this enables me to go to definitions defined by `await ...`

galaunay:
Jedi used to make the difference between "definitions" (like function and class definitions) and "assignments" (assigning abjects to variable names).

In this context, `y` does not have any definition but does have an assignment.
You can use `elpy-goto-assignment` to jump to it (see the [doc](https://elpy.readthedocs.io/en/latest/customization_tips.html?highlight=goto-assignment#alternatives-to-elpy-goto-definition)).

The new jedi API merged the two things into `goto`.
I was just waiting for the release of jedi 17.0 to make the change in Elpy.
So it should use `goto` soon.

Issue #1788, opened 2020-05-09
Elpy removes "ElDoc" indicator from modeline


ngleb:
# Summary

elpy modifies modeline. Setting variable `(setq elpy-remove-modeline-lighter nil)` does not help.

# Steps to reproduce

So basically I have two ways of reproducing the issue:
1) Config like below. Emacs started, I can see ElDoc indicator. If I open any *.py file, then ElDoc disappears.
2) Removing this line `  :mode ("\\.py\\'" . python-mode)` so that elpy is loaded at startup. Then `ElDoc` indicator disappears from the very beginning. Setting variable `(setq elpy-remove-modeline-lighter nil)` does not help.

This happens for ALL buffers, not just python-mode buffers.

## Elpy configuration in my init.el
```
(use-package python
  :mode ("\\.py\\'" . python-mode)
  :config
  (setq python-shell-interpreter "python"
        python-shell-interpreter-args "-i")
  (setq python-indent-guess-indent-offset nil))

(use-package elpy
  :after python
  :init
  (setq elpy-remove-modeline-lighter nil)
  :config
  (setq elpy-rpc-virtualenv-path 'current)
  (setq elpy-shell-echo-input nil)
  (delete 'elpy-module-highlight-indentation elpy-modules)
  (delete 'elpy-module-flymake elpy-modules)
  (add-hook 'elpy-mode-hook 'flycheck-mode)
  (elpy-enable))
```

Is it possible to completely disable any modifications to modeline except for adding "elpy'?

ngleb:
While this code checks this variable:
https://github.com/jorgenschaefer/elpy/blob/7e8181dbd6b900e6434218d2d5ccbac4c7615c41/elpy.el#L2748-L2756

It is not checked here:
https://github.com/jorgenschaefer/elpy/blob/7e8181dbd6b900e6434218d2d5ccbac4c7615c41/elpy.el#L3098

galaunay:
Absolutely right, I have merged the PR.

PR #1789, opened 2020-05-09
Make removal ElDoc from modeline conditional #1788

ngleb:
# PR Summary

Fix for #1788 

Instead of unconditionally removing ElDoc from modeline, use `elpy-modules-remove-modeline-lighter` to respect user's setting when `elpy-remove-modeline-lighter` set to `nil`.

6 unexpected failures, but they are not related to the proposed changes.

# PR checklist

## For new features only:

### DIFF
--- a/elpy.el
+++ b/elpy.el
@@ -3095,7 +3095,7 @@ and return the list."
   (pcase command
     (`global-init
      (require 'eldoc)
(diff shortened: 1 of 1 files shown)


galaunay:
Legitimate !

PR #1811, opened 2020-06-16
change elpy-rpc-pythonpath to support native-comp

tgbugs:
# PR Summary
The `.eln` files created by the native compile feature are not located in the same folder as the `.el` and `.elc` files they are derived from. This means that using `(locate-library "elpy")` to find the right folder for `elpy-rpc-pythonpath` breaks because it returns something like `"/home/tom/.emacs.d/elpa/elpy-20200527.2021/eln-x86_64-pc-linux-gnu-2ca71878e862259e/"` instead of `"/home/tom/.emacs.d/elpa/elpy-20200527.2021/"`. Using `(locate-library "elpy.el")` instead fixes the issue.

@AndreaCorallo Do you have an idea of the extent to which this will be an issue? In other words, how many codebases make use of `locate-library` in ways that will cause bugs that are similar to this? Probably just something to keep an eye on for the changelog for emacs 28, but might be worth starting some community awareness if it turns out to affect multiple packages.

# PR checklist

### DIFF
--- a/elpy-rpc.el
+++ b/elpy-rpc.el
@@ -96,7 +96,7 @@ for example), set this to the full interpreter path."
            (elpy-rpc-restart)))
   :group 'elpy)
 
-(defcustom elpy-rpc-pythonpath (file-name-directory (locate-library "elpy"))
+(defcustom elpy-rpc-pythonpath (file-name-directory (locate-library "elpy.el"))
   "A directory to add to the PYTHONPATH for the RPC process.
 
 This should be a directory where the elpy module can be found. If
--- a/elpy-rpc.el
+++ b/elpy-rpc.el
@@ -96,7 +96,7 @@ for example), set this to the full interpreter path."
            (elpy-rpc-restart)))
   :group 'elpy)
 
-(defcustom elpy-rpc-pythonpath (file-name-directory (locate-library "elpy.el"))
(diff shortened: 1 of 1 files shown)


AndreaCorallo:
Hi,  I'm not really sure but I think most packages use `load-file-name` for that.  

In fact I had to fake its value for compatibility and add `load-true-file-name` to have offer the possibility to obtain the real one.

Not sure if we could or should do something on the `locate-library` side.

tgbugs:
So the suggestion in this context would be to switch to use `(file-name-directory load-file-name)` since that will remain stable. I don't know enough about the decision to use `locate-library` in this context, so will have to wait for further input from one of the maintainers.

AndreaCorallo:
Well, more than a  suggestion was a comment :)

`load-file-name` will keep on pointing to the old directory because otherwise would break to many packages.  

If we think `locate-library` should work differently then what it does now it can certainly be discussed (perhaps on a bug on the Emacs bug tracker).

galaunay:
I am pretty happy with using the `load-file-name` solution in Elpy.
It looks more direct and stable than the `locate-library` alternative.

galaunay:
I took the liberty of modifying your PR following the above discussion.
 Thanks to both of you.

PR #1857, opened 2020-11-29
Use cl-lib when trying to use eval-sexp-fu-flash

dennisog:
# PR Summary

This is removing an implicit dependency on `cl' and updating it to use `cl-lib'. `cl-lib' is already a dependency.

# PR checklist

### DIFF
--- a/elpy-shell.el
+++ b/elpy-shell.el
@@ -417,7 +417,7 @@ active, additionally flashes that region briefly."
             (message "Sent: %s" code-on-first-line)
           (message "Sent: %s..." code-on-first-line))
         (when (bound-and-true-p eval-sexp-fu-flash-mode)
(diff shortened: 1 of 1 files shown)


gopar:
What emacs version do you have?

I tried with 26.1 and `multiple-value-bind ` is already provided in the global scope. Seems to be only an issue in emacs 27+

dennisog:
I am running some 27+ commit I grabbed off of the master branch. I seem to remember that `multiple-value-bind` is in `cl`, even in emacs26. 

Either way, since elpy already depends on `cl-lib`, we might as well use `cl-multiple-value-bind` here and be compatible with recent Emacsen.

galaunay:
legitimate,

looks fine to me, thanks

PR #1873, opened 2021-01-01
Update Class Snippets

gukgurgel:
# PR Summary

Address some problems of the snippet for creating classes, and they are:
- Unnecessary call of super() method inside __init__() of super classes;
- Call of super() method in Python2 syntax, which was updated in Python3

# PR checklist

- [x ] Commits respect our [guidelines](../CONTRIBUTING.rst)
- [x ] Tests are passing properly (see [here](https://elpy.readthedocs.io/en/latest/extending.html#running-tests) on how to run Elpy's tests)

## For new features only:

### DIFF
--- a/snippets/python-mode/class
+++ b/snippets/python-mode/class
@@ -1,13 +1,12 @@
 # -*- mode: snippet -*-
-# name: class(parent): ...
+# name: class: ...
 # key: class
 # group: Definitions
 # --
-class ${1:ClassName}(${2:object}):
-    """${3:Documentation for $1}
+class ${1:ClassName}:
+    """${2:Documentation for $1}
 
     """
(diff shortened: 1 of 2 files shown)


gopar:
Oh nice, I like the new snippet! I will say that even though py 2 has reached EOL, there are still plenty of ppl who will continue to use for a few years (At my previous company, there is no way we'll get off py2 until *years* later). Because of this, I'm towards leaving the `class` snippet alone and only include the `sclass`.

galaunay:
I like the idea of having `class` and `sclass` snippets.
In my opinion, we should use python3 syntax by default, and maybe providing a `py2class` (or else) snippet for python 2 users ?
what do you think ?

PR #1895, opened 2021-02-28
Pydantic integration with Jedi 18 fix.

gfederix:
# PR Summary
First steps in to integration pydantic.
Jedi16 support has been removed.
Fixed work of Jedi18.
Removed python2 support.

Code reviews and suggestions are welcome. :)

# PR checklist

- [ X] Tests are passing properly (see [here](https://elpy.readthedocs.io/en/latest/extending.html#running-tests) on how to run Elpy's tests)

## For new features only:


galaunay:
Oh my, that is a lot of commits :).

Is it ready to be tested and/or used ?

gfederix:
> Oh my, that is a lot of commits :).
> 
> Is it ready to be tested and/or used ?

I broke get_docstring. And this is only visible in emacs tests. ;-( Looks like other code works, but I need time to fix broken before merging. And probably write a regression test. I think I'll fix it in early April. 

I can also merge my commits into one, if it's convenient.

galaunay:
Well, I am afraid the test suite is not in a good state at the moment. Some tests fails even on the master branch, so nobody should focus too much on them ATM...
 We can squash the commits in due time, before merging,

PR #1902, opened 2021-03-28
Fix #1868: elpy-goto-definition breaks with python 3.9

### LINKED ISSUE #1868: elpy-goto-definition breaks on emacs 27.1 + python 3.9.1
# Summary

`elpy-goto-definition` was giving `nil` results on any function on emacs 27.1 with python 3.9.1

# Steps to reproduce

Call `elpy-goto-defintion` with emacs 27.1 and python 3.9.1 on any function. Upon a quick debug I saw the error from `elpy-rpc` as

```lisp
((message . "Object of type PosixPath is not JSON serializable")
 (code . 500)
 (data
  (traceback . "Traceback (most recent call last):
  File \"/home/joe/.emacs.d/elpa/elpy-20201115.1811/elpy/rpc.py\", line 104, in handle_request
    self.write_json(result=result,
  File \"/home/joe/.emacs.d/elpa/elpy-20201115.1811/elpy/rpc.py\", line 77, in write_json
    self.stdout.write(json.dumps(kwargs) + \"\\n\")
  File \"/lib64/python3.8/json/__init__.py\", line 231, in dumps
    return _default_encoder.encode(obj)
  File \"/lib64/python3.8/json/encoder.py\", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File \"/lib64/python3.8/json/encoder.py\", line 257, in iterencode
    return _iterencode(o, 0)
  File \"/lib64/python3.8/json/encoder.py\", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type PosixPath is not JSON serializable
")))
```

A simple fix in `rpc.py` like this works:

```python
    def json_defaults(self, x):
        if isinstance(x, Path):
            return str(x)

    def write_json(self, **kwargs):
        """Write an JSON object on a single line.

        The keyword arguments are interpreted as a single JSON object.
        It's not possible with this method to write non-objects.

        """
        self.stdout.write(json.dumps(kwargs, default=self.json_defaults) + "\n")
        self.stdout.flush()
```

galaunay:
Original solution from @akshaybadola here: https://github.com/jorgenschaefer/elpy/issues/1868

### DIFF
--- a/elpy/rpc.py
+++ b/elpy/rpc.py
@@ -10,6 +10,8 @@ See the documentation of the JSONRPCServer class for further details.
 import json
 import sys
 import traceback
+from pathlib import Path
+
 
 
 class JSONRPCServer(object):
--- a/elpy/rpc.py
+++ b/elpy/rpc.py
@@ -67,6 +67,10 @@ class JSONRPCServer(object):
             raise EOFError()
         return json.loads(line)
 
+    def json_defaults(self, x):
+        if isinstance(x, Path):
+            return str(x)
+
     def write_json(self, **kwargs):
         """Write an JSON object on a single line.
 
@@ -74,7 +78,7 @@ class JSONRPCServer(object):
         It's not possible with this method to write non-objects.
 
         """
-        self.stdout.write(json.dumps(kwargs) + "\n")
+        self.stdout.write(json.dumps(kwargs, default=self.json_defaults) + "\n")
         self.stdout.flush()
 
     def handle_request(self):


elpy/rpc.py
```diff
@@ -67,14 +67,18 @@ def read_json(self):
...
+
     def write_json(self, **kwargs):
         """Write an JSON object on a single line.
 
         The keyword arguments are interpreted as a single JSON object.
         It's not possible with this method to write non-objects.
 
         """
-        self.stdout.write(json.dumps(kwargs) + "\n")
+        self.stdout.write(json.dumps(kwargs, default=self.json_defaults) + "\n")
```
gopar, line 81: The docs for [`json.dumps` ](https://docs.python.org/3.9/library/json.html#json.dump) say that it should raise a `TypeError` if the object cannot be serializable, (In this case, all other objects that are not `Path` and are not primitive types will be encoded as `null` since `json_defaults` is implicitly returning `None`.

```python
>>> import json
>>> json.dumps(None)
'null'
>>> 
```
Also relevant [CPython source code](https://github.com/python/cpython/blob/63298930fb531ba2bb4f23bc3b915dbf1e17e9e1/Lib/json/encoder.py#L413)  that shows this will happen

I don't think this is the intended outcome?

akshaybadola: > The docs for [`json.dumps` ](https://docs.python.org/3.9/library/json.html#json.dump) say that it should raise a `TypeError` if the object cannot be serializable, (In this case, all other objects that are not `Path` and are not primitive types will be encoded as `null` since `json_defaults` is implicitly returning `None`.
> 
> ```python
> >>> import json
> >>> json.dumps(None)
> 'null'
> >>> 
> ```
> 
> Also relevant [CPython source code](https://github.com/python/cpython/blob/63298930fb531ba2bb4f23bc3b915dbf1e17e9e1/Lib/json/encoder.py#L413) that shows this will happen
> 
> I don't think this is the intended outcome?

Hi @gopar. Sorry I've been extremely busy. Yes, definitely that shouldn't be there. I had completely overlooked that and forgot to return the default value. It should be:

```python
def json_defaults(x):
    if isinstance(x, Path): 
         return str(x) 
    else: 
        return json.dumps(x)
```
See docstring of [default](https://github.com/python/cpython/blob/63298930fb531ba2bb4f23bc3b915dbf1e17e9e1/Lib/json/encoder.py#L160) in `JSONEncoder`.

Actually `pydantic` would be a good approach to go for it as it is primarily a parsing library. Though I think @gfederix's PR (See discussion in https://github.com/jorgenschaefer/elpy/pull/1895) requires a few fixes.

At this point I'm not sure which things to refactor, for example in `rpc.py`, as there are parallel changes in  https://github.com/jorgenschaefer/elpy/pull/1895 which are not yet merged. Any suggestions?

galaunay:
I merged #1916 instead of this one (the fix is the same, but the code looks cleaner).
We will revert it as soon as #1895 is ready !

PR #1916, opened 2021-05-15
Implement rudimentary compatibility with jedi 18.0

seppeljordan:
# PR Summary

This PR implements basic compatibility with jedi 18.0.  The issue was that the default json serializer used by python json.dumps could not handle `pathlib.Path` objects. This PR addresses the issue by extending the default JSONEncoder class. `pathlib.Path` objects are converted to str before encoding them to json.

# PR checklist

Test are not really passing BUT the tests were not passing before this change anyway. At least some more tests are passing now.

### DIFF
--- a/elpy/json_encoder.py
+++ b/elpy/json_encoder.py
@@ -1,0 +1,10 @@
+import json
+import pathlib
+
+
+class JSONEncoder(json.JSONEncoder):
+    def default(self, o):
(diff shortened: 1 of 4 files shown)


galaunay:
Duplicate of #1902.

#1895 will probably overwrite those modifications, but in the meantime, it will be nice to have jedi 0.18 working.
Thanks

Issue #1940, opened 2021-10-14
elpy fails to find python on fresh installs of Ubuntu 20.04


eanopolsky:
# Summary

On a fresh install of Ubuntu Server 20.04, python is not symlinked to python3, so elpy does not work out of the box. 

Systems that have been upgraded from previous Ubuntu releases to 20.04 do not seem to have this problem.

# Steps to reproduce

1. Install Ubuntu Server 20.04 using only the default options.
2. Install emacs from the package manager.
3. Install elpy from melpa using the instructions at https://github.com/jorgenschaefer/elpy/blob/master/README.rst
4. Start emacs.
5. Run `M-x elpy-config RET`.

What I expected to happen:
I expected elpy to set up an RPC virtual environment and eventually display the config interface.

What actually happened:
Elpy displayed `Cannot find executable 'python', please set 'elpy-rpc-python-command' to an existing executable.` and did not display the config interface.

# My configuration
## OS
Ubuntu Server 20.04.3.

## Result of `(elpy-config)`
`Cannot find executable 'python', please set 'elpy-rpc-python-command' to an existing executable.`

## Elpy configuration in my init.el
```
(use-package elpy
  :ensure t
  :init
  (elpy-enable))
```

# Workaround

Installing the `python-is-python3` package works around this issue.

PR #1941, opened 2021-10-14
fixes https://github.com/jorgenschaefer/elpy/issues/1940, tests

eanopolsky:
# PR Summary

On a fresh install of Ubuntu Server 20.04, python is not symlinked to python3, so elpy does not work out of the box. See https://github.com/jorgenschaefer/elpy/issues/1940 for details. Changes to elpy-rpc.el fix this.

Tests using python3 did not pass in my environment. Changes to scripts/test fix this.

# PR checklist

## For new features only:
N/A. This PR contains no new features.

### DIFF
--- a/scripts/setup
+++ b/scripts/setup
@@ -15,7 +15,7 @@ pip install -r requirements.txt --upgrade
 pip install -r requirements-rpc.txt --upgrade
 pip install -r requirements-dev.txt --upgrade
 # For python2
-if python -c 'import sys ; exit(sys.version_info < (3, 0))'
+if python -c 'import sys ; exit(not sys.version_info < (3, 0))'
 then
     pip install -r requirements-dev2.txt --upgrade
 fi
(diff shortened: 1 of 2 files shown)


scripts/setup
```diff
@@ -15,7 +15,7 @@ pip install -r requirements.txt --upgrade
 pip install -r requirements-rpc.txt --upgrade
 pip install -r requirements-dev.txt --upgrade
 # For python2
-if python -c 'import sys ; exit(sys.version_info < (3, 0))'
+if python -c 'import sys ; exit(not sys.version_info < (3, 0))'
```
gopar, line 18: This was not working before?

eanopolsky: Yes, it was not working before.

On my system, `python -c 'import sys ; exit(True)'; echo "exit(True) in python results in an exit code of $?"; python -c 'import sys ; exit(False)'; echo "exit(False) in python results in an exit code of $?"` prints:

```
exit(True) in python results in an exit code of 1
exit(False) in python results in an exit code of 0
```

However, bash interprets an exit code of 0 as true/success for flow control purposes, so it ends up running `pip install -r requirements-dev2.txt --upgrade` in a python 3 virtualenv, which broke a couple of tests.

gopar:
merged. thank you !

Issue #1949, opened 2021-12-03
elpy-black-fix-code does not use black config file


mwip:
# Summary
Opposed to `black` called from the command line, `elpy-black-fix-code` seems to ignore the `black` config (`~/.config/black` or `~\.black`). Is there a way to point to the config or maybe manually set the line length for the `elpy-black-fix-code`?

# Steps to reproduce
1. Create `black` config file at `~/.config/black
2. add the following to the file: 
```
[tool.black]
line-length = 100
target-version = ['py39']
```
3. open a python buffer, add very long line (that is between 88 (`black`'s line length default) and 100 (set via config file), e.g.
```
[str(bbbbbbbbbbbbbbbbbbbbbbb) for bbbbbbbbbbbbbbbbbbbbbb in range(bbbbbbbbbbbbbbbbbbbbbbbbbb)]
```
4. `M-x elpy-black-fix-code`

The long line should not be reformatted as it is below 100 columns long but still is. That is, `elpy` calling `black` on save does not seem to respect the default `black` config. 

# My configuration
## OS
Linux

## Result of `(elpy-config)`
```
Emacs.............: 27.2
Elpy..............: 1.35.0
Virtualenv........: None
Interactive Python: python 3.9.9 (/usr/bin/python)
RPC virtualenv....: rpc-venv (/home/user/.emacs.d/elpy/rpc-venv)
 Python...........: python 3.9.9 (/home/user/.emacs.d/elpy/rpc-venv/bin/python)
 Jedi.............: 0.18.1
 Autopep8.........: 1.6.0
 Yapf.............: 0.31.0
 Black............: 21.11b1
Syntax checker....: Not found (flake8)

```

## Elpy configuration in my init.el
```
     (use-package elpy
       :ensure t
       :config
       (elpy-enable))
     (add-hook 'elpy-mode-hook (lambda ()
				 (add-hook 'before-save-hook
					   'elpy-black-fix-code nil t)))

```

PR #1950, opened 2021-12-05
Use black for loading black config

jeremyheiler:
# PR Summary

Fixes https://github.com/jorgenschaefer/elpy/issues/1949

This changes how black configuration is loaded. Instead of looking for
`pyproject.toml` directly, use a black to load the configuration in the
same way the black CLI would. This allows for other configuration
files to be considered, such as `~/.black` or `~/.config/black`.

# PR checklist

### DIFF
--- a/elpy/blackutil.py
+++ b/elpy/blackutil.py
@@ -30,6 +30,14 @@ try:
         black = None
     else:
         import black
+        current_version = parse_version(black.__version__)
+        if current_version >= parse_version("21.5b1"):
+            from black.files import find_pyproject_toml
+        elif current_version >= parse_version("20.8b0"):
+            from black import find_pyproject_toml
+        else:
+            find_pyproject_toml = None
+
 except ImportError:  # pragma: no cover
     black = None
 
@@ -43,7 +51,10 @@ def fix_code(code, directory):
     # Get black config from pyproject.toml
     line_length = black.DEFAULT_LINE_LENGTH
     string_normalization = True
-    pyproject_path = os.path.join(directory, "pyproject.toml")
+    if find_pyproject_toml:
+        pyproject_path = find_pyproject_toml((directory,))
+    else:
+        pyproject_path = os.path.join(directory, "pyproject.toml")
     if toml is not None and os.path.exists(pyproject_path):
         pyproject_config = toml.load(pyproject_path)
         black_config = pyproject_config.get("tool", {}).get("black", {})


gopar:
Hey! Many thanks for the PR :)

Looks like `black.files` did not always exist until:
https://github.com/psf/black/commit/f2ea461e9e9fa5c47bb61fd72d512c748928badc (which was earlier this year)

The function `find_pyproject_toml` was instead in `black/__init__.py`. So for projects using an older version of black, this would break what we have for them :/

I do like the idea of letting black find the config file instead of whats happening in right now.

**ASK:**
Can you make a check that `find_pyproject_toml` exists in `black` and if it doesn't then import `black.files`? I feel this will less likely break things (ah the joys of trying **not** to break anyone's working elpy config :)

jeremyheiler:
> Can you make a check that `find_pyproject_toml` exists in `black` and if it doesn't then import `black.files`? I feel this will less likely break things (ah the joys of trying **not** to break anyone's working elpy config :)

Yes! I was actually thinking about this last night as I was falling asleep.

gopar:
Once the changes are implemented, I'll gladly merge!

jeremyheiler:
All set!

elpy/blackutil.py
```diff
@@ -30,6 +30,12 @@ def parse_version(*arg, **kwargs):
         black = None
     else:
         import black
+        current_version = parse_version(black.__version__)
+        if current_version >= parse_version("21.5b1"):
+            from black.files import find_pyproject_toml
+        elif current_version >= parse_version("20.8b0"):
```
gopar, line 36: Did `find_project_toml` not exist before this version?
If it didn't then we should have `find_project_toml = None` as an else branch so that we don't throw an error later down the road since `find_project_toml` won't be defined.

elpy/blackutil.py
```diff
@@ -30,6 +30,12 @@ def parse_version(*arg, **kwargs):
...
+            from black import find_pyproject_toml
+
 except ImportError:  # pragma: no cover
     black = None
```
gopar, line 40: Also put `find_project_toml = None` as a safety measure

gopar: actually ignore this one, if black isn't found, then the entire function won't run. Fine as is

elpy/blackutil.py
jeremyheiler: Yes, it was introduced about two years ago. I'll set it to None in the else block.

gopar:
Merged. Thank You!

posita:
This probably needs to be reverted or fixed. I'm pretty sure it is the root cause of #1955, which is currently being experienced in [`elpy-20211211.2248` on Melpa](https://melpa.org/#/elpy).

gopar:
@jeremyheiler do you have some time to look into this for a potential fix?

elpy/blackutil.py
```diff
@@ -43,7 +51,10 @@ def fix_code(code, directory):
     # Get black config from pyproject.toml
     line_length = black.DEFAULT_LINE_LENGTH
     string_normalization = True
-    pyproject_path = os.path.join(directory, "pyproject.toml")
+    if find_pyproject_toml:
+        pyproject_path = find_pyproject_toml((directory,))
```
posita, line 55: I _**think**_ this may be the root cause of #1955. It _**looks**_ like (at least with `black==21.12b0` that `find_pyproject_toml` will return `None` if it can't find `"pyproject.toml"` in `directory`.

Calling `os.path.exists(pyproject_path)` then equates to `os.path.exists(None)`, which will raise an exception:

``` python
>>> import os.path
>>> os.path.exists(None)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Library/Frameworks/MacPorts-20211130/Python.framework/Versions/3.9/lib/python3.9/genericpath.py", line 19, in exists
    os.stat(path)
TypeError: stat: path should be string, bytes, os.PathLike or integer, not NoneType
```

elpy/blackutil.py
```diff
@@ -43,7 +51,10 @@ def fix_code(code, directory):
...
+    else:
+        pyproject_path = os.path.join(directory, "pyproject.toml")
     if toml is not None and os.path.exists(pyproject_path):
```
posita, line 58: I think #1955 can be fixed by changing this line to:

``` python
if toml and pyproject_path and os.path.exists(pyproject_path):
```

I have hand-tested this, and it seems to cure the issue both in 0803e76e423286c92070eeeca0f2fe05df799b11 and in `master`, so I think it should do the trick.

posita:
> @jeremyheiler do you have some time to look into this for a potential fix?

Yup! See [this comment](https://github.com/jorgenschaefer/elpy/pull/1950/files/0803e76e423286c92070eeeca0f2fe05df799b11#r782690054) for a recommendation. I'm pretty sure that should do it.

You'll have to forgive me, past that, though. I don't have much experience with elisp or elpy's test mechanisms to formulate an automated way to repro this, so I think that's about as useful as I can be in this context (at least right now).

PR #1962, opened 2022-02-07
Fix bug#1917: elpy-shell-get-or-create-process fails for remote files

sam-s:
* elpy-shell.py (elpy-shell-get-or-create-process): Pass 'remote'
argument to 'executable-find' to enable remote executable search.

# PR Summary

# PR checklist

## For new features only:

### DIFF
--- a/elpy-shell.el
+++ b/elpy-shell.el
@@ -269,7 +269,7 @@ Python process. This allows the process to start up."
          (proc (get-buffer-process bufname)))
     (if proc
         proc
-      (unless (executable-find python-shell-interpreter)
+      (unless (executable-find python-shell-interpreter t)
         (error "Python shell interpreter `%s' cannot be found. Please set `python-shell-interpreter' to a valid python binary."
                python-shell-interpreter))
       (let ((default-directory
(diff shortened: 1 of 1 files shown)


elpy-shell.el
```diff
@@ -269,7 +269,7 @@ Python process. This allows the process to start up."
          (proc (get-buffer-process bufname)))
     (if proc
         proc
-      (unless (executable-find python-shell-interpreter)
+      (unless (executable-find python-shell-interpreter t)
```
gopar, line 272: Can you make this into a variable with a default of false?
Thank You!

sam-s: why would it ever need to be `nil`?
is the file is local, the check is local either way.
is the file is remote, the local check (with `nil`) is _pointless_ because the actual invocation of `python-shell-interpreter` will be done _remotely_ anyway.
thus using `nil` for the `executable-find` call here is always wrong.

sam-s:
Looks like the second argument to `executable-find` was added on 2018-06-20, and is not present in Emacs 26.2.

sam-s:
Now all the test failures are in python that I did not change - please advise!

gopar (approved):

gopar:
> Now all the test failures are in python that I did not change - please advise!

Ah don't worry about that. They've been failing for a while now.
Haven't had time to figure out the issue.

Thanks for making backwards compatible. I'll merge

PR #1969, opened 2022-03-18
Fix #1968 - test_should_get_docstring fails for Python 3.6+

### LINKED ISSUE #1968: RPCGetDocstringTests.test_should_get_docstring fails for Python 3.6+
# Summary

The test `RPCGetDocstringTests.test_should_get_docstring` in elpy/tests/support.py fails for Python 3.6+ because it looks for the "encoding" keyword argument for the docstring of `json.loads`. This keyword argument was deprecated in Python 3.1 and was later removed as a listed argument. I found [the commit where it was removed for Python 3.8+ ](https://github.com/python/cpython/commit/a8abe097c1165db25b429ca02a65c4f8acbc062b), but it appears that it was removed in 3.6 and 3.7 as well. 

The solution here is to remove the "encoding.\*, " portion of the `JSON_LOADS_REGEX` string. By removing the "encoding" keyword argument, this test will work for the versions of Python where this keyword was removed. For earlier versions of Python, the preceding greedy match regex (".*") will continue to match this keyword argument.

# Steps to reproduce
Run the following shell commands in the `elpy` source directory, with `virtualenv` using a recent build of Python version 3.6+.

> \> virtualenv ~/.virtualenvs/elpy
> \> source ~/.virtualenvs/elpy/bin/activate
> \> ./scripts/setup
> \> python -tt -m unittest discover elpy

drgillis:
This commit changes the regex string that matches against the arguments
of `json.loads` in the Python library. It removes the explicit
reference to the "encoding" keyword argument that was removed in
Python 3.6+.

# PR Summary
This commit fixes issue #1968. The test `RPCGetDocstringTests.test_should_get_docstring` in elpy/tests/support.py now passes for versions (3.6+) of Python where the "encoding" keyword has been removed from the signature of `json.loads`. 

Since the "encoding.\* " portion of the regex was preceded by a greedy wildcard, the new version of the regex will work regardless of whether this keyword is present in the `json.loads` signature.

# PR checklist

- [ X ] Commits respect our [guidelines](../CONTRIBUTING.rst)
- [ X ] Tests are passing properly (see [here](https://elpy.readthedocs.io/en/latest/extending.html#running-tests) on how to run Elpy's tests)

## For new features only:

### DIFF
--- a/elpy/tests/test_jedibackend.py
+++ b/elpy/tests/test_jedibackend.py
@@ -68,7 +68,7 @@ class TestRPCGetDocstring(RPCGetDocstringTests,
     def __init__(self, *args, **kwargs):
         super(TestRPCGetDocstring, self).__init__(*args, **kwargs)
         self.JSON_LOADS_REGEX = (
-            r'loads\(s.*, encoding.*, cls.*, object_hook.*, parse_float.*, '
+            r'loads\(s.*, cls.*, object_hook.*, parse_float.*, '
             r'parse_int.*, .*\)'
             )
 


gopar:
Thank you so much!

Issue #1966, opened 2022-02-24
Cannot interrupt process in inferior python if the process is started from an Elpy buffer


novusshu:
# Summary

This issue is driving me crazy. Simply put, if I start the interactive python with `run-python`, I can use `C-c C-c` which binds to `comint-interrupt-subjob` to interrupt current job (e.g., an infinite loop), and the process is still alive. However, I simply cannot use C-c C-c in an inferior python started from elpy mode python script (e.g., with `C-return`, or `C-c C-z`, etc. ).  Basically I cannot find a way to interrupt the process without kill the buffer.

I compared an inferior buffer started with `run-python` and with say, `C-return`, only to find that they have the exactly same minor modes and major mode (using `C-h m`). I did realize that Elpy has a slightly different way to start the inferior python based on the documentation for `elpy-shell-get-or-create-process`, in which it states that 

> ;; We cannot use `run-python` directly, as it selects the new shell
> ;; buffer. See https://github.com/jorgenschaefer/elpy/issues/1848

Yet, by comparing the code, I still cannot figure out why `C-c C-c` works in one way but not another.  Some help!!!

# Steps to reproduce
1. Open any .py file, write a simple infinite loop, e.g., 
```
while(1):
    print(1)
```
make sure your Elpy mode is enabled for the python buffer
2. Press Ctrl-<return>: an inferior buffer should start.
3. Switch to the inferior python buffer, press C-c C-c, and you will find yourself not able to interrupt the infinite loop. 

Now, if you change Step 2 to 2*:
2*. M-x run-python, 
Then in step 3, C-c C-c will do the job to terminate the loop.

# My configuration
## OS
MacOS 12.2.1

## Result of `(elpy-config)`
```
Emacs.............: 28.0.91
Elpy..............: 1.35.0
Virtualenv........: None
Interactive Python: python3 3.9.10 (/opt/homebrew/bin/python3)
RPC virtualenv....: rpc-venv (/Users/XXXX/.emacs.d/elpy/rpc-venv)
 Python...........: python3 3.8.9 (/Users/XXXX/.emacs.d/elpy/rpc-venv/bin/python3)
 Jedi.............: 0.18.1
 Autopep8.........: 1.6.0
 Yapf.............: 0.32.0
 Black............: 22.1.0
Syntax checker....: flake8 (/opt/homebrew/bin/flake8)

```

## Elpy configuration in my init.el
[...]

Issue #1948, opened 2021-11-24
C-d in pdb session causes shell exit


Yevgnen:
# Summary

Pressing <kbd>C-d</kbd> in a debugger session cause the REPL process exits. The <kbd>C-d</kbd> is bind to `comint-delchar-or-maybe-eof` as default. 

This happens to REPL process created by <kbd>C-c C-z</kbd>/`elpy-shell-switch-to-shell` but not <kbd>C-c C-p</kbd>/`run-python`.

This happens to a Python or iPython shell process.

This does not happen when entering `exit` and then RETURN in the debugger session.

# Steps to reproduce
1. Press <kbd>C-c C-z</kbd>(`elpy-shell-switch-to-shell`) in a `python-mode` buffer.
2. Enter `import pdb; pdb.set_trace()` in the REPL buffer then <kbd>ret</kbd>.
3. In the debugger session, press <kbd>C-d</kbd>.
4. The debugger session exits, then the REPL process also exits.

# My configuration
## OS
macOS 12.0.1 (21A559)

## Result of `(elpy-config)`
Sorry I can get this with the following messages after evaluating `(elpy-config)`

```
Elpy is creating the RPC virtualenv (’/Users/user/.emacs.d/elpy/rpc-venv’)
elpy-rpc--create-virtualenv: Doing vfork
```

## Elpy configuration in my init.el
```
(use-package elpy
  :custom-face
  (elpy-codecell-boundary ((t :underline t)))

  :diminish elpy-mode
  :hook (python-mode . elpy-enable)

  :init
  (setq elpy-modules '(elpy-module-sane-defaults
                       elpy-module-folding)
        elpy-shell-command-prefix-key "C-c C-e")

  :config
  (unbind-keys :map elpy-mode-map '("C-c C-d"))

  (use-package elpy-shell
    :straight elpy
    :init (setq elpy-shell-starting-directory 'current-directory
                elpy-shell-echo-output nil)))
```

drgillis:
The direct problem here is that Elpy connects to the Python shell via a pipe for Mac users. This was done because Python versions on MacOS that use the libedit version of `readline` (e.g. the default Mac Python versions) have issues operating with a pty connection (see e.g. #1671, #1550, #887). 

However, the good news is that many Python versions on Mac use GNU `readline` instead of the libedit one (e.g. versions installed by Homebrew or Anaconda). If you are using one of these versions, there is no need to need to use a pipe instead of a pty connection.  

[...]

Issue #1907, opened 2021-04-06
MacOS (only) use of pipe for process-connection type leads to EOF issues


jdtsmith:
I see that 'darwin (only) is [special-cased to run via a pipe](https://github.com/jorgenschaefer/elpy/blob/2203597e1254eba345d6873daa40c7b9d144931c/elpy-shell.el#L267), for reasons mentioned in #1671.  

This leads to some unfortunate behavior with iPython, in particular `(process-send-eof)` immediately closes the pipe with no further communication possible.  In normal iPython, you can close sub-shells like `ipdb` with `Ctrl-d`, and of course when you `Ctrl-d` the main iPython process, it prompts you (`Do you really want to exit ([y]/n)?`) before closing.  Unless there's still some special reason to run an interactive shell like `elpy`/`python-shell` through a pipe, sticking with the PTY seems more sensible.  

I had to include `--simple-prompt` when running ipython in pty mode to get iPython to notice our terminal's dumbness, but `Ctrl-d` then works entirely as expected and it seems slightly faster for small sends.

Issue #1838, opened 2020-09-29
Jupyter console displays "^G^G^G..." when sending buffer to shell


avigan:
# Summary

I have been encountering multiple issues with elpy, ipython and plotting with matplotlib for a while now, and following the documentation I have tried to switch to Jupyter console. However, when I use Jupyter console and send a region or buffer to the console it displays multiple `^G^G^G...` characters and the code is not executed. The fix for in IPython, which is to add `-c exec('__import__(\\'readline\\')')` to `python-shell-interpeter-args`, does not seem to work for Jupyter console. I get the following error:
```
[ZMQTerminalIPythonApp] CRITICAL | Unrecognized flag: '-c'
```
I guess there is maybe another way to pass the readline import? I have tried to type directly `exec('__import__(\'readline\')')` as the first command in the shell, but the problem remains the same. 

Any help would be much appreciated.

# My configuration
## OS
```
Mac OS X 10.14.6
```
## Result of `(elpy-config)`
```
Emacs.............: 28.0.50
Elpy..............: 1.34.0
Virtualenv........: None
Interactive Python: jupyter 4.6.3 (/Users/avigan/anaconda3/envs/default/bin/jupyter)
RPC virtualenv....: default (/Users/avigan/anaconda3/envs/default)
 Python...........: python 3.8.5 (/Users/avigan/anaconda3/envs/default/bin/python)
 Jedi.............: 0.17.1 (0.17.2 available)
 Rope.............: 0.17.0
 Autopep8.........: 1.5.4
 Yapf.............: 0.30.0
 Black............: Not found (20.8b1 available)
Syntax checker....: flake8 (/Users/avigan/anaconda3/envs/default/bin/flake8)
```
## Elpy configuration in my init.el
```
(elpy-enable)

(setq elpy-shell-send-region-or-buffer 0)
(setq elpy-shell-switch-to-shell 0)
(setq elpy-rpc-virtualenv-path 'current)

(setq python-shell-interpreter "jupyter"
      python-shell-interpreter-args "console --simple-prompt --no-confirm-exit"
      python-shell-prompt-detect-failure-warning nil)
(add-to-list 'python-shell-completion-native-disabled-interpreters
             "jupyter")
```

avigan:
My bad, the workaround proposed in https://github.com/jorgenschaefer/elpy/issues/1550#issuecomment-574512892 actually works for me: `(setq elpy-shell-echo-output nil)`

[...]

PR #1970, opened 2022-03-20
Fix #1867 - allow MacOS users to choose pty or pipe for Python shell

### LINKED ISSUE #1867: Problem with native completion on macOS, related to PR#1671
# Summary

`elpy-shell-switch-to-shell` gives the annoying warning:

```
Warning (python): Your ‘python-shell-interpreter’ doesn’t seem to support readline, yet ‘python-shell-completion-native-enable’ was t and "python3" is not part of the ‘python-shell-completion-native-disabled-interpreters’ list.  Native completions have been disabled locally.
```

This despite the `*Python*` buffer showing `python.el: native completion setup loaded`

`run-python` does not give the same warning.

I traced the difference in the two shell startup seqences to the following line in `elpy_shell.el`:
```
  (let* ((process-connection-type
          (if (string-equal system-type "darwin") nil t))  ;; see https://github.com/jorgenschaefer/elpy/pull/1671
```

If I change that to leave `process-connection-type` at `t` even on darwin, the warning goes away.  I tried the error case at the start of PR#1671 and could not reproduce that error, so the downside of having `process-connection-type t` on macOS seems to be gone.

I tried to find out why it fails with `process-connection-type nil` but that turned out to be beyond me.  I edebug-ed `python-shell-completion-native-get-completions` (in `python.el`) and ran `elpy-shell-switch-to-shell` and found out that it behaves differently depending on `process-connection-type`.

 - In the `nil` case, the redirect-buffer never receives any output after the call to `process-send-string`
 - In the `t` case the redirect-buffer is immediately filled with completion candidates starting with `_`, as expected.

I run Emacs from git, updated a couple of days ago.  Perhaps the problems with pty connection types have been resolved in Emacs 28 and it's time to revisit the existing workaround?

# Steps to reproduce

Visit a .py file, press `C-c C-z`, see the warning pop up.

# My configuration
## OS

macOS Mojave, 10.14.6

## Result of `(elpy-config)`
```
Elpy Configuration

Emacs.............: 28.0.50
Elpy..............: 1.35.0
Virtualenv........: None
Interactive Python: python3 3.9.1 (/usr/local/bin/python3)
RPC virtualenv....: rpc-venv (/Users/pok/.emacs.d/elpy/rpc-venv)
 Python...........: python3 3.9.1 (/Users/pok/.emacs.d/elpy/rpc-venv/bin/python3)
 Jedi.............: 0.17.2
 Rope.............: 0.18.0
 Autopep8.........: 1.5.4
 Yapf.............: 0.30.0
 Black............: 20.8b1
Syntax checker....: Not found (flake8)
```

## Elpy configuration in my init.el
```
  (use-package elpy
    :ensure t
    :init (elpy-enable)
    :config
    (setq elpy-rpc-python-command "python3")
    (setq python-shell-interpreter "python3"))
```

drgillis:
This change creates a custom variable that allows MacOS users to decide
whether to use a pipe or a pseudo-terminal (pty) to connect to the Python
shell when `elpy-shell-get-or-create-process` is called. Before this commit,
all Mac users are forced to use a pipe if the Python shell is created using
this function, even though this is not necessary for all users.

Some MacOS users experience issues connecting to the Python shell using a
pty. One of major problems is the inability to send greater than 1024
characters to the shell at one time. These issues are caused by versions of
Python that use the libedit version of Readline rather than the GNU
version. By default, MacOS does not have GNU Readline installed, so it instead
uses the libedit version for the versions of Python built into the OS. This
may also be the case for other versions installed on a Mac if GNU Readline was
not installed first. Using a pipe instead a pty alleviates all of these issues
at the expense of creating a few smaller issues (e.g. sending an EOF signal
into pdb causes the entire shell to exit).

However, many MacOS users use versions of Python that are built with GNU
Readline which means that the pty will work properly for them. It is
unnecessary for them to connect via a pipe. Additionally, users may
run into confusing behavior where the Python shell created via `run-python`
(which uses a pty unless `process-connection-type` is set to nil) acts
differently than one that is created via an Elpy command.

Since a pipe connection to the Python shell works for users with and without
versions of Python installed with GNU readline, this continues to be a sane
default for Mac users. Creating this custom variable provides an easy, if
somewhat difficult to discover, way to allow Mac users to decide which process
type they want to use.

In the future, it may be best to ask Mac users on their first usage of Elpy
whether to use a pipe or pty. Alternatively, this could be automatically
detected and a message could be sent to the Python interpreter to suggest that
they change this variable's value.

In addition to #1867, this change is also a fix for #1907 and #1948.

Other related issues and commits include: #887, #1550, #1838, and #1671.

# PR Summary
Several issues have been caused by forcing MacOS users to use a pipe instead of a pty to connect to the Python shell. The issues with using a pty (see #1550) only appear when Mac users use a version of Python that uses the libedit version of Readline rather than the GNU version (e.g. the default system installations of Python). It makes more sense to allow Mac users to choose whether to connect to the Python shell via a pty or via a pipe, so I created a custom variable that allows them to do this. 

I did not add any tests since writing good tests for this would be somewhat labor-intensive and this change only involves substituting a boolean-valued variable in the place of a nil value. **Edit to add:** I don't think I phrased this well. Since the main issues here depend on whether Python is built without an external library, this seems like this may be challenging to test with the current automated testing framework.

Issue #1966 is likely related as well.

# PR checklist

## For new features only:
- [X] The documentation has been updated

### DIFF
--- a/elpy-shell.el
+++ b/elpy-shell.el
@@ -175,6 +175,22 @@ the following functions:
   :type 'boolean
   :group 'elpy)
 
+(defcustom elpy-shell-darwin-use-pty nil
+  "Whether to connect to the Python shell through pty on MacOS.
+
+If nil, Elpy will connect to Python through a pipe. Any non-nil
+value will cause Elpy use a pseudo-terminal (pty) instead. This
+value should be set to nil when using a Python interpreter that
+uses the libedit version of Readline, such as the default MacOS
+Python interpreters. This value can be safely be set to true when
+using a version of Python that uses GNU Readline.
+
+This value is only used when `elpy-shell-get-or-create-process'
+creates a new Python process."
+  :type 'boolean
+  :group 'elpy)
+
+
 ;;;;;;;;;;;;;;;;;;
 ;;; Shell commands
 
@@ -280,7 +296,7 @@ REMOVE THIS when Elpy no longer supports Emacs 26."
 If SIT is non-nil, sit for that many seconds after creating a
 Python process. This allows the process to start up."
   (let* ((process-connection-type
-          (if (string-equal system-type "darwin") nil t))  ;; see https://github.com/jorgenschaefer/elpy/pull/1671
+          (if (string-equal system-type "darwin") elpy-shell-darwin-use-pty t))  ;; see https://github.com/jorgenschaefer/elpy/pull/1671
          (bufname (format "*%s*" (python-shell-get-process-name nil)))
          (proc (get-buffer-process bufname)))
     (if proc


gopar:
Thank You!

PR #1988, opened 2022-06-24
Replace find-tag-marker-ring with xref-push-marker-stack.

yatsky:
`find-tag-marker-ring` is obsolete as of Emacs 25.1 and I can't find it in 27.

# PR Summary

# PR checklist

## For new features only:


yatsky:
> Hey thanks for this! Can you make check so that it uses `find-tag-marker` on older versions and `xref-push-marker-stack` on newer versions? Thx!

Thanks I've added a check. If version < 25.1 (when  `find-tag-marker` was first marked as obsolete) then we use `find-tag-marker`, otherwise we use `xref-push-marker-stack`.

gopar (approved):

gopar:
Thank you!

PR #1995, opened 2022-08-17
Fix issue with eldoc when elpy-eldoc-show-current-function is nil

gmlarumbe:
# PR Summary
If using flycheck or some message utility while in a function,
having elpy-eldoc-show-current-function set to nil will cause
eldoc to delete any visible message in the echo area.

# PR checklist


gopar:
LGTM.
Merging.
Thanks!

PR #2014, opened 2023-05-10
Add support for tomli to parse pyproject.toml

mattusifer:
# PR Summary

# PR checklist

## For new features only:

### DIFF
--- a/elpy/blackutil.py
+++ b/elpy/blackutil.py
@@ -18,6 +18,11 @@ except ImportError:  # pragma: no cover
 import os
 
 try:
+    import tomli
+except ImportError:
+    tomli = None
+
+try:
     import toml
 except ImportError:
     toml = None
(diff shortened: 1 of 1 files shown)


gopar:
why remove the built in toml parser?
Also you are now forcing a dependency on tomli if we add this in

mattusifer:
This doesn't remove support for `toml`, it adds support for `tomli`. It's not forced, it is imported in the same "optional" way as `toml`:
```py
try:
    import tomli
except ImportError:
    tomli = None
```

Currently the `toml` lib is the only way elpy can parse pyproject.toml files, which makes it required, but `toml` isn't always in the environment. I think it would make sense to provide other options. Pip itself uses `tomli` to parse pyproject.toml files, and `tomllib` (a derivative of `tomli`) was recently added to the python standard library in 3.11 (so ultimately it will be best to just use that).

This PR isn't ready (yet). I didn't have `toml` in one of my venvs so I ran into this issue, still figuring out the best way to resolve it generally.

gopar:
Ah gotcha, completely misread the code. Ping whenever you want a review :)

steinitzu:
Is anything blocking this? Would love to have this as I currently have to patch my blackutil manually any time I update elpy.

Btw `tomli` is a dependency of black so imo fine to make it the only option. If you're using black, you have it installed.

PR #2015, opened 2023-05-22
Add support for python-ts-mode

gmlarumbe:
# PR Summary

`elpy-mode` requires current `major-mode` be a derived mode of `python-mode`. Emacs 29 bundles `python-mode` and `python-ts-mode` as derived modes of `python-base-mode`. This PR allows `elpy` to also be used with `python-ts-mode`.

# PR checklist

- [ ] Tests are passing properly (could not test it locally)


gopar:
How long have you been testing this? I don't use `python-ts-mode` so not sure if it breaks anything (highly unlikely though)but want to make sure it'll still be stable.

gmlarumbe:
> How long have you been testing this?

Just a couple of days. It seems to work fine but I can test it a bit more.

gopar:
> Just a couple of days. It seems to work fine but I can test it a bit more.


Yeah lets test for a ~month, to be on the conservative side.
Appreciate the PR

gopar:
Any issues @gmlarumbe ?
If not, i'll merge

gmlarumbe:
Hi @gomar,

I did not have much time to test it extensively. However it seems to work fine with both Emacs bundled `python-mode` and `python-ts-mode`, both derived from `python-base-mode`. I have not tried with MELPA `python-mode` but I assume it will work fine.

I also added a symlink for the yasnippet snippets of `python-base-mode` so that these could be used with `python-ts-mode`. I am not sure know how this is handled in Windows environments though.

gopar:
>  I also added a symlink for the yasnippet snippets of python-base-mode so that these could be used with python-ts-mode. I am not sure know how this is handled in Windows environments though.

I believe the correct way is to add [.yas-parents](https://joaotavora.github.io/yasnippet/snippet-organization.html#org7468fa9) and put `python-mode` in it.

gmlarumbe:
Hi @gopar ,

The .yas-parents for `python-ts-mode` in last force-push seems to work fine.

I also changed the enabling of elpy for `python-ts-mode` instead of for `python-base-mode` which in the end has the same effect.

gopar:
Merging. Not sure why the tests are failing there, but it should be unrelated

PR #2060, opened 2026-02-16
add choice for rpc for large buffers

kagetora66:
Prompts user for using rpc api if buffer is too large

# PR Summary
I regularly work with python files over 100Kb, I have found that the Emacs performance drop for large buffer sizes are rather negligible. So I think it could benefit from leaving the choice up to user if buffer is large.


gopar:
You can get around this by setting `elpy-rpc-ignored-buffer-size` to a higher number either in your `init.el` or as a `.dir-locals.el` entry. No need to ask :)

